PR #20, opened 2015-05-26
changed YawRate to Gyro for adding pitch and roll

aShinjiroUrata:
Please point out if there are better name.

### DIFF
--- a/vehicle_data/data_spec.html
+++ b/vehicle_data/data_spec.html
@@ -722,7 +722,7 @@
   </dl>
 </section>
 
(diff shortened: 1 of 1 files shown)


wonsuk73:
LGTM! 
@aShinjiroUrata minor comment: In 725 line, why do you use "Interface Acceleration" text?
@QingAn @acrofts84 @paulboyes Any review comment on this?

aShinjiroUrata:
>@aShinjiroUrata minor question: In 725 line, why do you use "Interface Acceleration" text?

That's my mistake. Fixed to "Interface Gyro" in another commit.

PR #23, opened 2015-05-26
mirrorTilt/mirrorPan to byte to allow signed values

acrofts84:
Updated mirrorPan and mirrorTilt to bytes to allow signed values


tripzero:
Looks good to me.

Issue #14, opened 2015-05-20
[data-spec] Vehicle Speed Units - Suggestion


acrofts84:
The units of the vehicle speed vary within the spec. The average speed from the trip clock, the top speed limit and the cruise control speed are in kilometers per hour whereas the vehicle and wheel speeds are given as meters per hour.

I can see why it has been done this way for usability, but I feel consistency may be beneficial. This then gives an argument for the use of SI units. I think this is a point that has been debated before, although regardless of whether SI units or not I would argue units should be consistent for the same measurements.

>6.2.1 Attributes
speed of type unsigned short, readonly
must return vehicle speed (Unit: meters per hour)

>6.3.1 Attributes
speed of type unsigned short, readonly
must return wheel speed (Unit: meters per hour)

>6.9.1 Attributes
averageSpeed of type unsigned short, readonly , nullable
must return average speed based on trip meter (Unit: kilometers per hour)

> 6.11.1 Attributes
speed of type unsigned short, readonly
must return target Cruise Control speed in kilometers per hour (Unit: kilometers per hour)

>9.5.1 Attributes
speed of type unsigned short, readonly
must return vehicle top speed limit (Unit: kilometers per hour)

djensen47:
- [ ] Get comments on which units should be used.
  - Argument for meters per hour is that it is the small unit available to avoid floats.
- [ ] Editors: update the doc to make units consistent.

paulboyes:
- [ ] Note in spec why certain units were chosen (such as meter per hour versus kilometer per hour)

djensen47:
As a point of reference, it looks like NASA JPL uses m/s for speed:

http://www2.jpl.nasa.gov/basics/units.php

PR #38, opened 2015-06-02
[data-spec] Vehicle speed units

acrofts84:
<a href='https://github.com/w3c/automotive/issues/14'>Issue 14</a>
I have added a note in the Conformance section with regards to units choice:

<pre>
Attribute units have been specified where appropriate. Integers have been
  used in favor of floating point values in order to preserve precision during 
  calculations (e.g. meters per hour instead of kilometers per hour).
</pre>

The vehicle speed values have also been updated throughout to meters per hour

### DIFF
--- a/vehicle_data/data_spec.html
+++ b/vehicle_data/data_spec.html
@@ -69,6 +69,11 @@
   specification MUST implement them in a manner consistent with the ECMAScript
   Bindings defined in the Web IDL specification [[!WEBIDL]], as this
   specification uses that specification and terminology.</p>
+  
+  <p>Attribute units have been specified where appropriate. Integers have been
+  used in favor of floating point values in order to preserve precision during 
+  calculations (e.g. meters per hour instead of kilometers per hour).</p>  
+  
 </section>
 
 <!---------------------------- Extending this API ----------------------------------->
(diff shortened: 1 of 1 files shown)


wonsuk73:
@acrofts84 I am not sure conformance section is right place for this phrase. Because typically conformance section of W3C spec does have very general descriptions as below. So I think chapter 6. might be one of appropriate place for this. What do you think?

"
2. Conformance
As well as sections marked as non-normative, all authoring guidelines, diagrams, examples, and notes in this specification are non-normative. Everything else in this specification is normative.

The key words must, must not, required, should, should not, recommended, may, and optional in this specification are to be interpreted as described in [RFC2119].

This specification defines conformance criteria that apply to a single product: the user agent that implements the interfaces that it contains.

Implementations that use ECMAScript to implement the APIs defined in this specification must implement them in a manner consistent with the ECMAScript Bindings defined in the Web IDL specification [WEBIDL], as this specification uses that specification and terminology.
"

acrofts84:
Hi Wonsuk, 
Apologies for the delay, I've been on holiday!
I agree with you, I wasn't sure that the conformance section was the most appropriate. However I still wanted this to be noted at a high level, since it applies to the speed units in Section 9 as well as Section 6. The same point also applies to millimetres in the size configuration within Section 5. 
Although I agree that Section 6 is the most relevant section. The other options would be add it to the Introduction section or the top of each relevant section? The Introduction is probably most appropriate to avoid repeating the point. Does that sound appropriate?

wonsuk73:
@acrofts84 Thanks for feedback. I think this phase need to be written as a editor's note, because it's for editors of the spec, not a reader. Therefore how about adding the sentence to section 5, 6, 9 repeatedly. Then for the last version we could remove those :) What do you think?

@QingAn @paulboyes any feedback?

QingAn:
@acrofts84  @wonsuk73  I agree that the sentence can be added to each relevant section as editor's note.

wonsuk73:
@acrofts84 If you agree with my proposal, could you make pull request for that?

Issue #5, opened 2015-05-11
[data_spec] nullable and non-nullable attribute definition


QingAn:
In data spec, http://www.w3.org/2014/automotive/data_spec.html   
for the multiple kinds of attributes defined in different interfaces, some attributes are defined as nullable, like fuelType, which is nullable, in FuelConfiguration Interface (http://www.w3.org/2014/automotive/data_spec.html#idl-def-FuelConfiguration), and VehiclePowerMode, which is non-nullable, in VehiclePowerModeType Interface (http://www.w3.org/2014/automotive/data_spec.html#idl-def-VehiclePowerModeType)

We need a clear explanation why such attributes are set in this way.

djensen47:
Action items

- [ ] Check with Kevron (@tripzero) and Justin (@jsbach86) why some are nullable and notnullable.
- [ ] Document the guidelines

tripzero:
The general rule I think we've been following is that if the interface is provided, it MUST implement the non-nullable items otherwise it's optional.  Many of the non-nullable attributes are found in interfaces with only one attribute (ie VehicleSpeed).  Logically it doesn't follow to even offer the interface (ie make it available as an attribute on "vehicle") if it has no attributes.

PR #39, opened 2015-06-02
Document the guidelines

QingAn:
nullable and non-nullable attribute


QingAn:
#5

wonsuk73:
@QingAn I am not sure this sentence should be added to intro section. Because I think it's meaningful for the spec editors. So how about adding this as an editor's note in appropriate place of the spec?

QingAn:
@wonsuk73 That make sense. How about set this sentence as "note" and still add to intro section. Please check the updated version.

Issue #17, opened 2015-05-21
[data-spec] Steering Wheel Left - Ambiguous


acrofts84:
The steeringWheelLeft attribute name feels very ambiguous to me. Would this be better off as a Zone? Or a less ambiguous term e.g. leftHandDrive or driverSideLeft. Although I don't know how well these translate.

<pre>5.7.1 Attributes
steeringWheelLeft of type boolean, readonly , nullable
must return true if steering wheel is on left side of vehicle
</pre>

djensen47:
- [ ] Ask previous editors for their comments on this.

paulboyes:
- How to handle cars with no steering wheel?
- How to handle single seat car?
- Center steer wheel?
- Other interfaces?  Such as head unit location?  What similar interfaces are we missing?

tripzero:
You could make it steeringWheelOrientation with just a zone attribute.

PR #41, opened 2015-06-15
[data-spec] SteeringWheelConfiguration Update: Steering Wheel Location

acrofts84:
As discussed in <a href="https://github.com/w3c/automotive/issues/17">Issue 17</a>.
In order to remove ambiguity and allow for any orientation of steering wheel (and null in the case of no steering wheel) this interface has been updated from:

<pre>
steeringWheelLeft of type boolean, readonly , nullable
must return true if steering wheel is on left side of vehicle
</pre>
to
<pre>
steeringWheelLocation of type Zone, readonly , nullable
must return the location on the steering wheel within the vehicle
</pre>

### DIFF
--- a/vehicle_data/data_spec.html
+++ b/vehicle_data/data_spec.html
@@ -305,8 +305,8 @@
   information about a vehicle.</p>
 
   <dl title="interface SteeringWheelConfiguration : VehicleCommonDataType" class="idl">
-     <dt>readonly attribute boolean? steeringWheelLeft</dt>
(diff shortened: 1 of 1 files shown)


wonsuk73:
It makes sense to me!

@QingAn @aShinjiroUrata @tripzero do you have any opinion on this?

QingAn:
I agree!

aShinjiroUrata:
I think it is good!

Issue #32, opened 2015-05-27
[data-spec] 7.10 - Diagnostic Interface


gbrannonaaa:
For Diagnostic interface in 7.10, some additional data elements that could be considered

- diagnostic trouble codes - array
	- including 213 EPA mandated codes and OEM-specific, as available

wonsuk73:
@QingAn @acrofts84 @aShinjiroUrata Could you share your view on this issue? I would like to have more feedback for this item.

aShinjiroUrata:
If this is interface for DTC(diagnostic trouble code) of OBD2, maybe better to have this interface, I suppose.

gbrannonaaa:
Our view if that is seems very odd to omit this very important piece of information for the owner of the vehicle.  Someone using the API could utilize the DTC to inform the consumer about issues that are going on with their car.  The business group limited spec to only include information about the MIL lamp.  This is similar to the doctor telling you that you are sick, that you have been for a while, but not telling you why or what part of your body is the problem.  The spec should not be limited to only malfunctions that turn on the MIL, but any DTC.  The API user can determine what they wish to do with the data.

wonsuk73:
@aShinjiroUrata Thanks! @gbrannonaaa Thanks for additional info!. I have couple of questions!

1) When does this TroubleCode interface be used? We also have VehicleInterfaceError Interface in Vehicle Information Access API. What is different between them in the aspect of a role?

2)  In addition, we have simple example code [1] in the intro section of API spec. So could you make some example code how we could use TroubleCode interface?

[1] http://rawgit.com/w3c/automotive/master/vehicle_data/vehicle_spec.html#introduction

acrofts84:
Some cars do have methods for accessing their DTCs but they seem fairly obfuscated and I think there is an important reason for this. Exposing this data to a wider audience is more likely to increase customer dissatisfaction. 
A DTC is meant to be logged on the vehicle if the driver <i>could</i> notice an issue. If it is important, then a warning light will come on. If they don't notice, then there is no issue as far as they are concerned. If they do notice, there will be a physical symptom which they can investigate.
If you have use cases where we can equally obfuscate this information from a casual driver and then aid the more savvy drivers then I'm happy to listen. Although in the age where cars are essentially computers with wheels there are probably going to be more potential DTCs, and less that the savvy driver can do about them without the appropriate software.

gbrannonaaa:
[...]

PR #43, opened 2015-06-17
Update data_spec.html

gbrannonaaa:
To address issue #32

### DIFF
--- a/vehicle_data/data_spec.html
+++ b/vehicle_data/data_spec.html
@@ -1022,6 +1022,34 @@
   </dl>
 </section>
 
+
+<!-- Interface  TroubleCode -->
+<section>
+
+    <h3><a>TroubleCode</a> Interface</h3> <p>The <a>TroubleCode</a> interface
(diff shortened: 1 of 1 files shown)


wonsuk73:
@gbrannonaaa, TroubleCodeType have three types like current, pending and history in the proposal. What does 'pending' type mean?

gbrannonaaa:
It is one on three defined types of codes by SAE standard.  OEMs will have no issues understanding the meaning.

wonsuk73:
@gbrannonaaa, I understood which types are from. By the way I am still curious about meaning of type 'pending' ;)

gbrannonaaa:
OBD II Pending Codes are also referred to as “continuous monitor” and “maturing codes”. An intermittent fault will cause the computer to store a pending code in memory. If the fault does not recur within a certain number of cycles, the code will be cleared from memory. If the fault recurs a specific number of times, the code will then mature into a diagnostic trouble code (DTC).

Issue #33, opened 2015-05-30
[data-spec] DriveMode and DrivingMode


aShinjiroUrata:
<pre>
6.26 DrivingMode Interface
8.6 DriveMode Interface
</pre>
I felt similar name and may be confusing. 
If `DrivingMode` is not commonly used name, how about `DrivingStatus` or some other name?
(I'm not sure about suitable naming.)

aShinjiroUrata:
It looks the name '6.26 DrivingMode' is selected to align with '6.27 NightMode', '6.28 StartStopMode' and so I guess this naming is reasonable from this view point.

On the other hand, '8.6 DriveMode' looks natural name.

Since I don't come up with better alternative for 'DrivingMode', I'd like to keep this as is if other people don't have other opinion.

acrofts84:
I agree that they are very similar. Would 'VehicleInMotion' or 'MotionStatus' be a more appropriate name instead of 'DrivingMode'?

aShinjiroUrata:
Thanks for the input.
My concern is just the names were too similar and so, both `VehicleInMotion`, `MotionStatus` are OK to me. 
For me, I like shorter one. 
If there is no other opinion, I will create pull request with `MotionStatus`.

aShinjiroUrata:
As pull request #46 "renamed DrivingMode to DrivingStatus" already merged, I close this issue.

PR #46, opened 2015-07-03
renamed DrivingMode to DrivingStatus

aShinjiroUrata:
I have made a pull request for issue #33 .
Previously I wrote that I would make a pull request to rename to `MotionStatus` by @acrofts84 's advice.
However, `MotionStatus` sounds like showing the car is moving or not and I felt it doesn't fit well.
Even if a car is not moving, the car should be considered as in driving state in some cases (such as temporary stop at signal), in terms of drivers distraction avoidance.
So I changed the name to `DrivingStatus`.
What do you think?

### DIFF
--- a/vehicle_data/data_spec.html
+++ b/vehicle_data/data_spec.html
@@ -377,8 +377,8 @@
         <dd>MUST return VehicleSignalInterface for accessing <a>WheelTick</a></dd>
         <dt>readonly attribute VehicleSignalInterface buttonEvent</dt>
         <dd>MUST return VehicleSignalInterface for accessing <a>ButtonEvent</a></dd>
-        <dt>readonly attribute VehicleSignalInterface drivingMode</dt>
-        <dd>MUST return VehicleSignalInterface for accessing <a>DrivingMode</a></dd>
+        <dt>readonly attribute VehicleSignalInterface drivingStatus</dt>
+        <dd>MUST return VehicleSignalInterface for accessing <a>DrivingStatus</a></dd>
         <dt>readonly attribute VehicleSignalInterface nightMode</dt>
         <dd>MUST return VehicleSignalInterface for accessing <a>NightMode</a></dd>
       </dl>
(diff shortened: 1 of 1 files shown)


acrofts84:
A lot of the driver distraction guidelines I've read mention vehicle motion. JAMA [1] and NHTSA [2] both talk about motion in terms of what to display to the driver. Hence my suggestion of MotionStatus.

<i>"A display system's functions that are not presumed to be used by the driver during driving operation shall be inoperative by the driver while the vehicle is in motion."</i> JAMA [1]

<i>"System functions not intended to be used by the driver while driving should be made inaccessible for the purpose of driver interaction while the vehicle is in motion."</i> NHTSA [2]

However, I agree there is an advantage of keeping it broader than that. The OEM is open to implement it however they wish, be it determined by speed/gear/steering wheel movement/hand brake status. This extra level of flexibility may be an advantage to the interface's implementation, and as it is documented "may combine several other data types" then I'll agree that DrivingStatus is probably suitable. Sorry for sending us on a tangent!

[1] http://www.jama-english.jp/release/release/2005/jama_guidelines_v30_en.pdf
[2] http://www.nhtsa.gov/staticfiles/rulemaking/pdf/Distraction_NPFG-02162012.pdf

aShinjiroUrata:
Thanks for the explanation. I understand the reason to choose the words `VehicleInMotion` and `MotionStatus` and these are appropriate words for this purpose.

>However, I agree there is an advantage of keeping it broader than that. 
>The OEM is open to implement it however they wish, be it determined by speed/gear/steering
> wheel movement/hand brake status. This extra level of flexibility may be an advantage to the
> interface's implementation, and as it is documented "may combine several other data types"
> then I'll agree that DrivingStatus is probably suitable.

I agree. Then, for the flexibility advantage, I'll keep the `DrivingStatus` pull request as it is.
By the explanation, now I think any of these three words is OK for the name and not strongly stick to `DrivingStatus`.

wonsuk73:
Thanks @aShinjiroUrata and @acrofts84. I agreed with `DrivingStatus` too!

PR #48, opened 2015-07-07
removal of etc from spec

MelcoGot:
Please review the removal of etc from spec. Beware of limitations that these changes could indicate.

### DIFF
--- a/vehicle_data/data_spec.html
+++ b/vehicle_data/data_spec.html
@@ -145,7 +145,7 @@
       <h2>Configuration and Identification Interfaces</h2>
 
       <p>Interfaces relating to vehicle configuration and identification including: make, type size, transmission
-      configuration, identification numbers, etc...</p>
+      configuration and identification numbers</p>
 
       <dl title="partial interface Vehicle" class="idl">
         <!-- identification and configuration types: -->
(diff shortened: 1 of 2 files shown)


acrofts84:
I don't believe this has introduced any limitations because there is always a word/phrase which states that the list is not exhaustive ('including', 'e.g' or 'such as'). I also agree that it reads better as a result of this change.

I have a couple of comments for readability, but that is it. Changing this:
<pre>Driver able to see fuel such as economy, consumption, level and range.</pre>
With
<pre>Driver able to see fuel status such as economy, consumption, level and range.</pre>

And this (I realise this isn't anything to do with your proposed change):
<pre>and is given an snapshot</pre>
With:
<pre>and is given a snapshot</pre>

wonsuk73:
@MelcoGot, could you update your pull request to have comments from @acrofts84 ?

peterMelco:
Since I don't have write access - how to do with the comments from acrotft ?

wonsuk73:
@peterMelco you are different guy with @MelcoGot. Right? just checking! :)  Can I ask which write access do you mean?

wonsuk73:
This pull request made conflict to be merged. So it has been done via Pull request #57 and #58

peterMelco:
Hi !
Not sure…I am probably both…
/pw

Issue #31, opened 2015-05-27
[data-spec] 7.8 BatteryStatus Interface


gbrannonaaa:
For battery status in 7.8, some additional data elements that could be considered

- battery purpose (starting, hybrid-traction, electric-traction, other)
- cold cranking amps rating
- battery location (under hood, under seat, trunk, other)
- battery temperature 
- battery type (flooded lead-acid, enhanced flooded lead-acid, AGM, AGM/Spiral, lithium, capacitor, other)
- cranking voltage
- charging system voltage

PR #54, opened 2015-07-21
[data_spec] BatteryStatus Interface Extension

acrofts84:
Hi @gbrannonaaa 

Thank you for the update, it looks comprehensive. I've created a pull request for <a href ="https://github.com/w3c/automotive/issues/31">Issue 31</a> in the w3c/automotive repository, as the pull request only existed in your fork.

It does pose the question of which datatype we should be using for temperatures. If we do not want to use floats for speed, then should we take a similar tact with temperature? At the moment we use short, byte, long and float to represent temperatures in the spec. Should we standardise this? @wonsuk73, @QingAn do you have an opinion?

### DIFF
--- a/vehicle_data/data_spec.html
+++ b/vehicle_data/data_spec.html
@@ -999,6 +999,76 @@
      <dd>MUST return battery current (Unit: amperes).</dd>
      <dt>readonly attribute Zone? zone</dt>
      <dd>MUST return Zone for requested attribute</dd>
+     
+
+     <dt>readonly attribute DOMstring? batteryType</dt>
+     <dd>MUST return battery type (SLI (starting, lighting, ignition), stop-start, hybrid relay, hybrid traction)</dd>
+     <dt>readonly attribute DOMstring? batteryChemistry</dt>
+     <dd>MUST return battery chemistry (Flooded lead acid, AGM, enhanced flooded, lithium)</dd>
+     <dt>readonly attribute short? batteryCCACapacity</dt>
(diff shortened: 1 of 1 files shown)


acrofts84:
The units are not stated for all attributes. Since many of them are temperatures/voltages is it a good idea to change the preamble the following to avoid repetition, and then explicitly state other units:

<pre>
The BatteryStatus interface provides information about the state of a vehicle's battery. All temperatures are represented in celsius and all voltages are represented in volts.
</pre>

wonsuk73:
@acrofts84 Sorry for late reply. Using preamble for temperatures/voltages is good to me. Starting point  of chapter 7 would be appropriate I guess. What do you think?

@QingAn What do you think?

tripzero:
Can we specify somewhere we are using SI units (with the exception of
Celcius instead of Kelvin)?  Then specify any other exceptions as needed?

wonsuk73:
@tripzero Your proposal make sense to me. I guess we can add the sentence to after current last sentence of Intro chapter. @acrofts84 @QingAn @gbrannonaaa What do you think?

wonsuk73:
It is required to make pull request for following up below items 
- Adding a sentence to intro chapter of data spec about it's basically use SI units for whole document
- Adding a sentence to starting point of chapter 7 regarding to the exception of Celcius instead of Kelvin

gbrannonaaa:
It seems using SI units for the whole document makes the most sense if everyone is in agreement with that approach.

PR #58, opened 2015-08-04
[vehicle spec] editorial - removal of etc from spec

wonsuk73:
editorial - removal of etc from spec


wonsuk73:
This Pull request is to cover the updated parts for vehicle spec from Pull request #48 (It made conflict to merge)

PR #60, opened 2015-08-10
[data-spec] SI units/temperature units

acrofts84:
I have added a sentence in the Introduction: 

<i>"Unless otherwise stated, SI units should be used where applicable for all attributes."</i>

And a comment within the Maintenance interface: 

<i>"Temperature attributes within this interface should be specified in degrees Celsius."</i>

As per comments on https://github.com/w3c/automotive/pull/54

### DIFF
--- a/vehicle_data/data_spec.html
+++ b/vehicle_data/data_spec.html
@@ -79,7 +79,8 @@
 <!-- Introduction -->
 <section class="informative">
   <h2>Introduction</h2>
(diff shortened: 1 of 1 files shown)


acrofts84:
I considered this, but I didn't want to create a list of exceptions, remembering that another exception, the speed units in meters per second, is specified as an editors note as per Issue #14. Although there doesn't appear to be other units which are consistent across the spec and not SI.

Would this be a better solution:

<i>"Unless otherwise stated, SI units should be used where applicable for all attributes. Notable exceptions for this are temperature which shall be specified in Celsius and speed which shall be specified in meters per hour."</i>

PR #67, opened 2015-09-18
[data-spec] Adding parking sensors interface

acrofts84:
An interface to return the distance to the nearest object using the parking sensors. Provides a distance attribute, an active attribute and a zone attribute.

<pre> <code>
active of type boolean, readonly , nullable
must return true if parking sensors are currently in use
distanceToObject of type short, readonly , nullable
must return the approximate distance to the nearest object from the parking sensor. If no objects are sensed this should return null (Unit: millimeters)
zone of type Zone, readonly , nullable
must return Zone for requested attribute
</code></pre>

### DIFF
--- a/vehicle_data/data_spec.html
+++ b/vehicle_data/data_spec.html
@@ -1781,6 +1781,8 @@
         <dd>MUST return VehicleSignalInterface for accessing <a>Alarm</a></dd>
         <dt>readonly attribute VehicleSignalInterface parkingBrake</dt>
         <dd>MUST return VehicleSignalInterface for accessing <a>ParkingBrake</a></dd>
+        <dt>readonly attribute VehicleSignalInterface parkingSensors</dt>
+        <dd>MUST return VehicleSignalInterface for accessing <a>ParkingSensors</a></dd>
         <dt>readonly attribute VehicleSignalInterface parkingLights</dt>
         <dd>MUST return VehicleSignalInterface for accessing <a>ParkingLights</a></dd>
       </dl>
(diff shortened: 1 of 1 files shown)


wonsuk73:
@acrofts84, Recently new car has a lot of parking sensors in back and front sides as well. Could your proposal can cover all of sensors where are back and front in any density?

wonsuk73:
@aShinjiroUrata @QingAn Do you have any comment on this PR?

aShinjiroUrata:
Hi! basically this is OK to me.

This sensor is used only at parking time?
If sensor information can be used while car is running,
 then names like `distanceSensors`, `proximitySensors` would be other option.

Bt the way,
>@acrofts84, Recently new car has a lot of parking sensors in back and front sides as well. Could >your proposal cover all of sensors where are back and front in any density?

as personal opinion, OEM can put as many sensors as they like and I think it is not possible to 
 cover such various sensors of various models of cars by current `Zone` method. 
But I think, it is not this `parkingSensors` interface's problem but `Zone` system's limitation.

wonsuk73:
@aShinjiroUrata, Thanks for comments!
@acrofts84 I will merge your proposal to the spec with editor note for indicating that it might need to revisit the name of 'parkingSensors'. And this PR also has a conflict. So I will make an alternative PR with  editor's note.

wonsuk73:
This PR has been replaced with PR #70 because of conflict.

acrofts84:
Thanks @wonsuk73 @aShinjiroUrata.

PR #71, opened 2015-10-14
[data-spec] Add axes system for the Gyro interface

acrofts84:
Add information to clarify the polarity used in the axes system for the gyro: ISO 8855 Z-up standard. Also updated steering angle and mirror tilt to reflect this.

This has been verified as the accepted standard within raw CAN data.


wonsuk73:
@acrofts84, Can I ask why we need to follow the ISO 8855 Z-up standard? Unfortunately I cannot see ISO 8855 Z-up standard. What do you think guys, @aShinjiroUrata @QingAn?

acrofts84:
It is a standard that defines a vehicle axes system, so the rotations are supplied with a consistent polarity. I added the note on the standard so people could look it up further in case my notes were not clear enough, and it adds a justification to the use of the axis system. Both Paul and I have confirmed with CAN signals and within Vehicle Dynamics that this is the appropriate standard to follow. 

Is there any disadvantage to including this in the spec?

wonsuk73:
@acrofts84 Okay, No problem to add! I just want to know shortly. Thanks :)

PR #74, opened 2015-11-23
Changed data type for speed from unsigned short to unsigned long

jsbach86:
Unit for speed was changed from km/h to m/h, so data type should have been changed accordingly as well.
The range of unsigned short is up to 65535 usually, so couldn't represent vehicle speed completely.


wonsuk73:
Thanks! @jsbach86. It makes sense!

PR #89, opened 2016-08-30
Clarify that RESTful Web services are currently out of scope

acrofts84:


wonsuk73:
Looks good to me! It's more clear than previous.

PR #92, opened 2016-09-19
applying some of meeting results

wonsuk73:
- made clear sentence for “The mechanisms will include WebSockets and REST”
- removed “The APIs developed by this group may use a different security model from the traditional browser security model.” and “Hence, it is conceivable that it is used/implemented without using the Vehicle Signal Server Specification.”
- fixed typos


tguild (approved):

aShinjiroUrata:
Excuse me for asking basic question.
Do we need to keep REST mechanism?
- In my understanding, RESTful API is http based mechanism which is good for one shot request such as get(), set() but not good for subscribe(). 
  This is because, in subscribe(), server have to push data to client without client's request. 
- As personal opinion, most people would be satisfied with Websocket based server spec. I don't have an idea when someone wants to use REST spec.
- REST based spec will be considerably different from Websocket based spec. If websocket mechanism is enought, it might be redundant effort to create REST spec. 

I'd like to know if there is special reason that we should keep REST mechanism.

peterMelco:
Hi Urata-san!

I basically agree with you - in our current thoughts on what we will use in out reference implementation we dont have a case for REST as of now.

 However, I think that Adam and Kevin made a point of keeping the REST api ?

Br
Peter Winzell

---

Från: Shinjiro Urata [notifications@github.com]
Skickat: den 20 september 2016 10:34
Till: w3c/automotive
Ämne: Re: [w3c/automotive] applying some of meeting results (#92)

Excuse me for asking basic question.
Do we need to keep REST mechanism?
-   In my understanding, RESTful API is http based mechanism which is good for one shot request such as get(), set() but not good for subscribe(). This is because, in subscribe(), server have to push data to client without client's request.
-   As personal opinion, most people would be satisfied with Websocket based server spec.
-   REST based spec will be considerably different from Websocket based spec. If websocket mechanism is enought, it might be redundant effort to create REST spec.

I'd like to know if there is special reason that we should keep REST mechanism.

peterMelco:
Hi All!

I also have a basic question that I think we answered in F2F in Portland - I have forgotten (as usual). The introduction of a request id is motivated by what ? Why is it not sufficient to use the session id to identify the request per session ? I think that Adam and Kevin had a good explanation for it ...

Br
Peter Winzell

---

Från: Shinjiro Urata [notifications@github.com]
Skickat: den 20 september 2016 10:34
Till: w3c/automotive
Ämne: Re: [w3c/automotive] applying some of meeting results (#92)

Excuse me for asking basic question.
Do we need to keep REST mechanism?
-   In my understanding, RESTful API is http based mechanism which is good for one shot request such as get(), set() but not good for subscribe(). This is because, in subscribe(), server have to push data to client without client's request.
-   As personal opinion, most people would be satisfied with Websocket based server spec.
-   REST based spec will be considerably different from Websocket based spec. If websocket mechanism is enought, it might be redundant effort to create REST spec.

I'd like to know if there is special reason that we should keep REST mechanism.

wonsuk73:
@aShinjiroUrata Thanks for reasonable comment. I agreed. I will replace 'REST' with appropriate words like advanced protocol :-) What do you think?

wonsuk73:
@tguild I had some of confused problems with Github desktop app. but anyway I sent a new PR for applying discussion results to the charter. Could you review and apply? Thanks!

tguild:
Thanks Wonsuk

aShinjiroUrata:
Hi Peter-san, Wonsuk-san, thanks for comment.

> @aShinjiroUrata Thanks for reasonable comment. I agreed. I will replace 'REST' with appropriate words like advanced protocol :-) What do you think?

Then you think it's better to include some more abstract concept of protocol than just limit only to Websocket, right? Why you think this is necessary?
Excuse me but I don't come up with appropriate replacement word.

I think options will be
- delete REST and just limit the scope to Websocket
- keep the word REST
- use concrete words like HTTP instead of REST
  ( In WoT document, they include concrete protocols such as HTTP/CoAP/MQTT/Websocket.
  http://w3c.github.io/wot/current-practices/wot-practices.html#protocol-bindings )

From feasibility point of view, 'subscribe' looks realizable even on HTTP via '(Long-)Polling/Chunking/Streaming', although I don't know if these are good realization method in terms of easy-to-use, stability, error-handling.

So regarding feasibility, keeping 'REST' in scope can be OK, but still I feel this might be a redundant work...

aShinjiroUrata:
Regarding the sentence in the charter document, 

> The mechanisms will be firstly based on WebSockets and it could be extended to apply advanced protocols in the future.

I think this is very good. We can concentrate on WebSocket for the initial version.

wonsuk73:
@aShinjiroUrata Thanks for review!

PR #100, opened 2016-10-17
Add unsubscribe all action

acrofts84:
Introduce the unsubscribeAll action to avoid the use of a magic number (previously used subscriptionId = 0).
- This uses the existing unsubscribe mechanisms with a different action
- As a result, subscriptionId is now nullable


wonsuk73:
@acrofts84 It looks good to me. @aShinjiroUrata @drkevg What do you think?

aShinjiroUrata:
Hi, this looks good to me.

PR #113, opened 2016-11-15
Version Numbering for WVSS

acrofts84:
I haven't explicitly given a version number to this document. I presume the W3C gives high level versioning to specs where appropriate.


acrofts84:
Will create new pull request to satisfy this change

Issue #107, opened 2016-10-23
discrepancy between title and short title


dret:
currently the title is "Vehicle Signal Server Specification" and the short title is "vehicle-information-service" (signal server vs. information service). to me, service makes much more sense (no consumer is interested in the appliance providing a service), but whatever it is, it should be consistent.

rstreif:
Per Working Group Meeting: rename title to "Vehicle Information Service Specification"

dret:
looks good to me!

rstreif:
In addition per Working Group Meeting: name the companion document specifying the API "Vehicle Information API Specification"

drkevg:
Updated name of server spec to be 'Vehicle Information Service Specification' and for references to client spec to 'Vehicle Information API Specification

PR #114, opened 2017-01-04
Update title to Vehicle Information Service Spec

drkevg:


drkevg:
Implemented in response to issue #107. Changes reviewed by Adam (Crofts)

PR #115, opened 2017-01-05
Use DOMString instead of Strings

acrofts84:
Ensure all strings are DOMStrings i.a.w. WebIDL and to ensure the document is consistent


acrofts84:

PR #116, opened 2017-01-05
Note to state that JSON is the data format used throughout

acrofts84:
"The data format used for client/server communication within this specification is JSON encoded strings, however other data formats may be supported in future versions of the specification."


acrofts84:

PR #117, opened 2017-01-05
WebSocket security token diagram clarification

acrofts84:
The diagram in Figure 3 does not use the exact data structures from the specification, but it is demonstrative and simpler than including the actual structures which would need additional fields, such as timestamps and requestIds. For simplicity I think the diagram should remain the same, but it is important to note that it is conceptual.


peterMelco:
agree

acrofts84:

PR #118, opened 2017-01-05
Clarify behaviour for unsubscribeAll action

acrofts84:
If the unsubscribeAll action is used, but there are no active subscriptions, this will still return a success response as the subscriptions will be in the state desired by the user.


acrofts84:

PR #120, opened 2017-01-05
Additional Errors - read_only and too_many_attempts

acrofts84:
Added read_only as an error for setting signals.
Added too_many_attempts to prevent dictionary attacks for authentication


acrofts84:
Reviewed with @drkevg

PR #122, opened 2017-01-09
Add Vehicle Signal Client Spec

pkinney:
As discussed during our call and CES F2F, this is a specification for the Vehicle Signal Client based on the guidelines laid out and documented at the Portland F2F.  The goal is to define a specification that respects current best-practices in front-end development as well as promote the concepts of the VISS on the client side.


wonsuk73:
@pkinney Thanks for proposal! Let's start to make client api spec with this one!

PR #128, opened 2017-01-11
requestId and subscriptionId should be strings

acrofts84:
This allows them to be numeric or UUIDs etc.


wonsuk73:
It makes sense to me.

PR #129, opened 2017-01-11
Note on wwwivi as a hostname

acrofts84:
Discuss merits of using 'wwwivi' as a standardised hostname. Do we want a more generic hostname, not just for IVI but also encompassing WoT standards?


wonsuk73:
Naming is always difficult because people has their own triangle for the object. To me wwwivi is okay to me but I doesn't have strong intention for that. Let's try to gather more comments from members.

PR #131, opened 2017-01-17
Add JSON Schema Definitions to Each Interface

acrofts84:
Added a heading to each section showing the existing Web IDL, followed by a heading for JSON Schema and the equivalent JSON Schema

### DIFF
--- a/vehicle_data/vehicle_information_service.html
+++ b/vehicle_data/vehicle_information_service.html
@@ -1876,7 +1876,7 @@
     <section>
       <h3>Server Side Filtering</h3>
       <p>'<a>Filters</a>' may be specified to enable Server side filtering
-      to be used in order to throttle the demands of subscriptions on
+      to be used to throttle the demands of subscriptions on
       the server. This may enable the reduction of traffic if the
       developer has received a 429 - Too Many Requests error
       message. Server side filters are extensible, allowing for further filtering mechanisms. Support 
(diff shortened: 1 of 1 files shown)


acrofts84:
Thank you both, I have integrated your recommendations.

rstreif:
Thank you @acrofts84. I am ok with merging.

wonsuk73:
@acrofts84 Sorry for late reply because I had a little big long vacation :-)
Can I ask why doesn't remove the WebIDL definition? If it is to check the matches in detail during migration from WebIDL to JSON schema, that is okay. But...

If there are two definitions for VISS. I think it could make a confusion for the readers because WebIDL and JSON schema use different data types like DOMString, DOMTimeStamp and Any. So my proposal is only to use JSON schema for VISS in this spec.

acrofts84:
@wonsuk73, no problem. I hope you enjoyed it :).

The reason to keep the WebIDL is it is more human readable.  I agree that the data types may lead to confusion. I can see the argument for both keeping and removing the WebIDL, so was planning to raise this issue and hopefully come to a consensus in the call on 21st Feb.

PeterWinzell:
I think we also had an argument that if someone would like to have a browser implementation of the spec it would be helpful with WebIDL defintions. I agree with @wonsuk73  that it could and mayabe actually should  be removed.

wonsuk73:
@acrofts84 @peterMelco As you know in case of WebIDL that is typically used for javascript API which is primitively supported by the browser. It is not appropriate for server side APIs like VISS. It should be used our client spec. What do you think?

peterMelco:
@wonsuk73  I completely agree with you.

acrofts84:
As discussed over email, I think the best way forward is to merge this request and the removal/replacement of Web IDL can be the subject of a future pull request.

Issue #132, opened 2017-01-17
What fomat of json should be returned by getVSS()?


aShinjiroUrata:
One possible format option should be the one defined in VSS spec which includes 
 a lot of information such as type, description, min, max, unit, etc. 
https://github.com/GENIVI/vehicle_signal_specification/blob/develop/vss_rel_1.json

Another option should be something more light weight one which includes only necessary information.

I guess both may be possible but I suppose better to define in the spec.

rstreif:
My suggestion would be to define a flag as part of the API that provides selection of the format. In its simplest form that could be a boolean flag for a short form that just provides the signal paths and the ID and a long form that provides all information.

aShinjiroUrata:
>My suggestion would be to define a flag as part of the API that provides selection of the format.

Thanks for the comment!
This looks practical solution.

Reason why I raised this issue is, I first though returning simple JSON like below.
But in this case, it may be difficult to tell a node is a 'leaf' or a 'branch with no children'.
So I thought 'type' information may be necessary.

```
{Signal:
  {DriveTrain:
    {InternalCombustionEngine:
      {RPM:
        {type:UInt16
         min:0,
         max:20000,
         id:54,
         unit:'rpm'}
      }
    }
  }
 Attribute:{},
}
```

On the other hand, using VSS json is like below
 and this include necessary information.
But I felt this might be too much such as, 
 I don't want 'children' node and other not necessary information for client.
(Of course, not a big problem. 
 It will be OK by just ignoring 'children' node and other attributes. 
 Or no need to include properties like 'description' in returned json. 
 I guess Data size will be not a big problem as well, since getVSS() may not
 be used too frequently.)
```
{Signal:
  {type:branch,
   description:'xxx'
   children:
    {DriveTrain:
      {type:'branch',
       description:'aaa',
       children:
        {InternalCombustionEngine:
          {type:'branch',
           description:'aaa'
           children:
            {RPM:
              {type:UInt16
               min:0,
               max:20000,
               id:54,
               unit:'rpm'}
            },
          }
        },
      }
    }
  }
}
```

And I wanted to ask someone has any idea or preference on this.

rstreif:
[...]

PR #133, opened 2017-01-24
Clarify the getVSS method

acrofts84:
Introduce a data structure in the example and add further explanation

### DIFF
--- a/vehicle_data/vehicle_information_service.html
+++ b/vehicle_data/vehicle_information_service.html
@@ -815,20 +815,48 @@
       };
       </pre>
 
-    <p>The following data flow example shows a request for the signal structure within the Signal.Body branch containing signals related to the vehicle body.</p>
+    <p>The following data flow example shows a request for the signal structure within the Signal.Drivetrain.InternalCombustionEngine.RPM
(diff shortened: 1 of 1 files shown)


rstreif:
This looks good to me and reflects the full VSS. It's ok to be merged from my side. I still would like to entertain the idea raised in #133 adding a parameter to getVSS to specify either this verbose format or a more compact format. But that can be another PR.

aShinjiroUrata:
Hi @acrofts84 , thank you very much for the update!
Just for minor point, I suppose the request path should be "Signal.Drivetrain.InternalCombustionEngine.RPM" than "Signal.Body" in this case, isn't it?

Except for this point, the change looks fine to me.

aShinjiroUrata:
Hi @acrofts84 , thank you for the fix!
This PR is OK to me.

acrofts84:
Thank you both. I've updated the request path. I'll add to the discussion regarding the verbose mode in issue #132.

wonsuk73:
@acrofts84 @rstreif @aShinjiroUrata Thanks for contributions. It looks good to me as well. So it just  merged.

PR #138, opened 2017-02-16
Provide a more substantial example

acrofts84:
Added getVSS, subscriptionIds, and a generic onMessage handler

### DIFF
--- a/vehicle_data/vehicle_information_service.html
+++ b/vehicle_data/vehicle_information_service.html
@@ -200,6 +200,9 @@ vehicle.onmessage = function(event){
       setReqHandler(msg);
       break;
     case "subscribe":
+      subscriptionSetupHandler(msg);
+      break;
+    case "subscription":
       subscriptionHandler(msg);
       break;
     case "unsubscribe":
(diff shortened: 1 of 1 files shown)


aShinjiroUrata:
Hi Adam-san,

Thanks for example improvement.
I think I found minor mistake.

In EXAMPLE3, this looks you expect `subscriptionNotification` comes in to `subscriptionHandler()`.
However, in the current definition, `subscriptionNotification` doesn't have `action` property and
 so, doesn't come in to 'subscriptionHandler()'.

In this case, I guess it might be better to add `action`(==subscribe) to `subscriptionNotification`
 than fix EXAMPLE. What do you think?

Except for this point, I think it's OK to merge.

acrofts84:
Hi Urata-san. Thank you for your review and for raising my error. I have raised your comment in #141, and hopefully we will conclude this on the call this evening

aShinjiroUrata:
I had been thinking `action` for `subscriptionNotification` object should be `subscribe`.

However, in #141 discussion, the `action` string for `subscriptionNotification`
 was decided as `subscription` not `subscribe`.

So, I think #138 example should be modified to use `action: subsctiption`.

acrofts84:
Thank you Urata-san. I've updated the example to include the "subscription" action, and will raise another pull request to add the subscription action to the document itself.

acrofts84:
I have merged this with #154

PR #139, opened 2017-02-16
Clarify subscription behaviour

acrofts84:
Explicitly state that the tree unsubscribes from an entire subscription, you cannot just do just a subset


wonsuk73:
@acrofts84, It makes sense to me!

Issue #141, opened 2017-03-13
Subscribe repsonse inconsistent


peterMelco:
**Section: 8. Message Structure 
Cat: Subscribe**

When subscribing to signals the action attribute is not part of the return JSON message, this is inconsistent . The inconsistency makes testing ugly.

Suggestion: add "action", "subscribing" to respond message.

acrofts84:
Hi Peter. Thanks for the comments. I agree that it is ugly. Urata-san mentioned this in #138, where I had assumed that the subscriptionNotification would return the action: "subscribe".

I agree that this is ugly and we should include something here. Maybe action = "subscription" or "subscriptionNotification"

rstreif:
AR: Resolve the issue as suggested by this issue and issue #138: add action: "subscription" to the returned JSON.

peterMelco:
Sounds good.

acrofts84:
Action "subscription" added and merged

PR #154, opened 2017-03-24
Add "subscription" action and update example

acrofts84:
Integrate changes suggested in #138 and #141.
- Add "subscription" action to subscriptionNotification and subscriptionNotificationError
- Provide a more substantial example, including the above change


aShinjiroUrata:
Adam-san

Thank you very much for the update.
I think this is OK to merge.

acrofts84:
Thanks Urata-san

PR #159, opened 2017-03-30
Fixed the wrong message and added description for examples in the unsubscribe part

wonsuk73:
- removed subscriptionId from response for unsubscribeAll example
- revised descriptions for each example for better understanding


wonsuk73:
@acrofts84 Could you review this?^^ It's simple though :-)

Issue #147, opened 2017-03-21
F) 4. W3CClient() > Functions: subscribe() detail


aShinjiroUrata:
Seprarated from #137

====================
@aShinjiroUrata wrote at 2017 Feb22

F) 4. W3CClient() > Functions: subscribe()

In VISS, ranges, interval, minimum changes are handled in filters.
To make it aligned with VISS, the options should have only filter and onValue. Then interval, onChange should be specified inside filter.
What do you think?
If using same definition of filter, then VIAS library need to do nothing excpet just passing filter string to VISS' subscribe method.

====================
@pkinney wrote at 2017 Mar8

F) Agreed, that we should directly reference the filter spec from VISS here if possible so as not to duplicate anything.

====================
@aShinjiroUrata wrote at 2017 Mar21

Not yet modified as described above.

aShinjiroUrata:
In the current VIAS, subscribe will be executedd like follows.
```
// Current VIAS subscribe
client.subscribe(
    // path
    'Signal.Drivetrain.Transmission.Speed', 
    //options
    {   'interval' : 100,
        'onChange' :true,
        'filters' :{  // defined in VISS
            'interval':100,
            'range':{'above':100, 'below':200},
            'minChange':10
        },
        'onValue' : function(_val, _subscription, _client) { console.log("value = "+_val); }
    },
    // onSubscribe callback
    function(_err, _subscription) { /* subscribe succeed */ }
);
```
Current VIAS subscribe's problems is ...
- 'interval' can be configured as member of 'filters' and as member of 'options'.
- concept of 'onChange' and 'minChange' are not the same but similar
- 'options' and 'filters' looks repetitive and complex

I suppose, better to have "VIAS subscribe's parameter" usage is mostly same with "VISS subscribe's parameter", because it's simpler and less confusing.
And moreover, VIAS subscribe's implementation become much easier.

I'd like to change VIAS subscribe as below.
```
client.subscribe(
    // path
    'Signal.Drivetrain.Transmission.Speed', 
    //filters (defined in VISS)
    {
        'interval':100,
        'range':{'above':100, 'below':200},
        'minChange':10
    },
    // onValue callback
    function(_val, _subscription) { console.log("value = "+_val); },
    // onSubscribe callback
    function(_err, _subscription) { /* subscribe succeed */ }
);
```

What do you think?

### 2017/4/3 added:
I have changed mind a little.
There should be no need to have two callback funcs
 and I have deleted one as below.

[...]

PR #167, opened 2017-04-03
For issue #147, updated subscribe()'s interface definition

aShinjiroUrata:


aShinjiroUrata:
I think this change is not so controversial nor problematice,
 and also because of time limitation I have merged this PR.
If you find any problem, please raise as new github issue.

PR #168, opened 2017-04-04
Revert "Fixed the wrong message and added description for examples in the unsubscribe part"

wonsuk73:
Reverts w3c/automotive#159


wonsuk73:
Sorry. I think I had merged only PR #159 to the repo. but one more PR had been merged at the same time. So I had reverted two PRs! I will try to do PR one by one soon. @acrofts84

PR #169, opened 2017-04-04
Updated the unsubscribe example part 

wonsuk73:
- Removed subscriptionId from response for unsubscribeAll example
- Revised descriptions for each example for better understanding

### DIFF
--- a/vehicle_data/vehicle_information_service.html
+++ b/vehicle_data/vehicle_information_service.html
@@ -1906,7 +1906,7 @@ function subscriptionHandler(msg){
 </pre>
 
       <h4>Examples</h4>
-	Unsubscribe from a single subscription.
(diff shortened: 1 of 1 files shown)


wonsuk73:
@acrofts84, Could you check below JSON example for unsubscribeAll action? I has removed ' "subscriptionId": null ' from receive message. Because that info is optional and doesn't need for unsubscribeAll.  Does it make sense to you?

	client -> {
		"action": "unsubscribeAll",
		"requestId": "3468"
	}

	receive <- {
		"action": "unsubscribeAll",
		"requestId": "3468",
		"timestamp": <DOMTimeStamp>
	}

acrofts84:
Thanks Wonsuk :)

Issue #171, opened 2017-04-04
Handling optional property in JSON schema


wonsuk73:
Below is the JSON schema for vssRequest in the current draft and vssRequest consists of 'action', 'path', 'requestId' properties. 'path' is only optional property. Below schema defines all of three properties including 'path' as a mandatory to be described. Then 'path' could have a 'null' value.

I think it is not the best way for handling optional property in JSON schema.
~~~~
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "title": "VSS Request",
    "description": "Request metadata describing the potentially available VSS tree",
    "type": "object",
    "required": ["action", "path", "requestId"],
    "properties": {
        "action": {
            "enum": [ "getVSS" ],
            "description": "The identifier for the getVSS request",
        },
        "path": {
            "$ref": "#/definitions/path",
            "type": ["string", "null"]
        },
        "requestId": {
            "$ref": "#/definitions/requestId"
        }
    }
}
~~~~
My proposal is like this. VSS Request object consists of three properties but mandatory are 'action' and 'requestId'. So in case of 'path' that could be used when the case is required to use 'path'.
~~~~
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "title": "VSS Request",
    "description": "Request metadata describing the potentially available VSS tree",
    "type": "object",
    "required": ["action", "requestId"],
    "properties": {
        "action": {
            "enum": [ "getVSS" ],
            "description": "The identifier for the getVSS request"
        },
        "path": {
            "$ref": "#/definitions/path"
        },
        "requestId": {
            "$ref": "#/definitions/requestId"
        }
    }
}
~~~~
@acrofts84, What do you think? I will prepare PRs for applying this to the whole of draft.

rstreif:
@wonsuk73, Does this mean if the path is omitted then the entire VSS tree is returned?
I a sense an empty path seems to be implied by `"type": ["string", "null"]`. Consequently, if path can be empty or null then it makes only sense to make it optional.
Makes sense to me.

wonsuk73:
@rstreif,
Yes. As you think omitting 'path' property means to ask  the whole of VSS file. When the client want to ask only the some part of VSS tree, client should use 'path' property to indicate the part he want. I think for better understanding it would be good to add examples to getVSS part for both of cases. If we have a consensus on this, I will make a PR for applying it.

rstreif:
@wonsuk73 
[...]

PR #173, opened 2017-04-04
Revised JSON schema for subscribeRequest and unsubscribe interface #171

wonsuk73:
Regarding issue #171 it had been revised for 
- JSON schema of subscribeRequest in subscribe interface
- JSON schema of unsubscribeRequest, unsubscribeSuccessResponse and unsubscribeErrorResponse in unsubscribe interface

### DIFF
--- a/vehicle_data/vehicle_information_service.html
+++ b/vehicle_data/vehicle_information_service.html
@@ -1579,11 +1579,11 @@ function subscriptionHandler(msg){
     "description": "Allows the client to subscribe to time-varying signal notifications on the 
       server.",
     "type": "object",
-    "required": ["action", "path", "filters", "requestId"],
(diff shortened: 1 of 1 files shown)


rstreif:
@wonsuk73, makes logically sense since the JSON schema allows a null value for subscriptionId, which means it can be made optional.
The only clarification that might be necessary, is what an empty subscriptionId actually means. Does it mean unsubscribing from all events?

wonsuk73:
@rstreif Thanks for your opinions!

Empty subscrptionId is used only for unsubscribeAll action! for removing all of subscription at one time. I think if that is used in any case related with unsubscribe action, that might be wrong.

acrofts84:
Thanks @wonsuk73. I agree, this is clearer and more logical

PR #176, opened 2017-04-10
Revised JSON Schema Definitions for base datatypes

wonsuk73:
- Added "$schema" and "title" to the schema
- Removed "null" from the type of "filters"
- Fixed errors from JSON validator

### DIFF
--- a/vehicle_data/vehicle_information_service.html
+++ b/vehicle_data/vehicle_information_service.html
@@ -700,10 +700,13 @@ function subscriptionHandler(msg){
 <p>The definitions within this section describe the datatypes referenced within the JSON Schema VISS interfaces.</p>
 <pre>
 {
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "Base datatype definitions for the JSON schema VISS interfaces",
     "definitions": {
         "action": {
+	    "type": "string",
             "enum": [ "authorize", "getVSS", "get", "set", "subscribe", "subscription", "unsubscribe", "unsubscribeAll"],
-            "description": "The type of action requested by the client or delivered by the server",
+            "description": "The type of action requested by the client or delivered by the server"
         },
         "requestId": {
             "description": "Returned by the server in the response and used by the client to 
@@ -728,7 +731,7 @@ function subscriptionHandler(msg){
         "filters": {
             "description": "May be specified in order to throttle the demands of subscriptions 
               on the server.",
-            "type": ["object", "null"],
+            "type": "object",
             "properties": {
                 "interval": {
                     "description": "The subscription will provide notifications with a period 
@@ -749,7 +752,7 @@ function subscriptionHandler(msg){
                         "description": "The subscription will provide notifications when the value 
                           is greater than or equal to this field's value.",
                         "type": "integer"  
-                      },
+                      }
                     }                   
                 },
                 "minChange": {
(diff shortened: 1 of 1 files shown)


wonsuk73:
@acrofts84 and @drkevg,  Could you review this?

acrofts84:
Hi @wonsuk73. 

Thank you for these updates. Filters, and a couple of other things, are 'nullable' in the WebIDL definition. My understanding was that this is not analogous to omittable, therefore in order to be consistent between the JSON Schema and the WebIDL, null should be an accepted type in JSON Schema. Is my understanding correct here? 

I do not disagree that omitting these values is better than including them as 'null', but I'd like to ensure this is consistent and use the correct definitions.

wonsuk73:
@acrofts84, Thanks for the comment.
Now probably I got your point exactly. So far I was confused the meaning of '?' character in WebIDL. I understood as optional for '?' character in WebIDL. But that is not. It means 'nullable' like you said. Sorry for that. But we need to think about again whether the usage of both description language, WebIDL and JSON schema, for VISS spec is really good for real world :-) Because WebIDL has a limitation to describe JSON message format I guess.

Below are all of interfaces that nullable type is used in VISS spec. I think it is better to use optional property than nullable property for those. because even the case property doesn't have a value to describe, that property should be described in JSON message. I don't think it's good shape. But unfortunately I couldn't look for the way to define optional attribute in the interface definition of WebIDL syntax. In addition this is also one of mismatch issue between WebIDL and JSON schema when we use both description language to define JSON message structure in the spec.

In conclusion, if we still would like to use nullable type in WebIDL for those property, then JSON schema definition part should be rollbacked to previous version. Or if we would like to only use JSON schema, then all of WebIDL definition should be replace with other human readable shape. :-)

What do you think?

#1. 'path' property in vssRequest interface 
~~~~
interface vssRequest {
    attribute Action     action;
    attribute DOMString? path;
    attribute DOMString  requestId;
};
~~~~

#2. 'filter' in subscribeRequest interface
~~~~
interface subscribeRequest {
    attribute Action    action;
    attribute DOMString path;
    attribute object?   filters;
    attribute DOMString requestId;
};
~~~~

#3. 'subscriptionId' in  unsubscribeRequest, unsubscribeSuccessResponse and unsubscribeErrorResponse interfaces
~~~~
interface unsubscribeRequest {
    attribute Action     action;
    attribute DOMString? subscriptionId;
    attribute DOMString  requestId;
};

interface unsubscribeSuccessResponse {
    attribute Action       action;
    attribute DOMString?   subscriptionId;
    attribute DOMString    requestId;
    attribute DOMTimeStamp timestamp;
};

interface unsubscribeErrorResponse {
    attribute Action       action;
    attribute DOMString?   subscriptionId;
    attribute DOMString    requestId;
    attribute Error        error;
    attribute DOMTimeStamp timestamp;
};
~~~~

acrofts84:
Hi Wonsuk, 
The WebIDL is frustratingly close. I don't know if using TypeScript notation is more appropriate?
<pre>
interface vssRequest {  
    action: Action;
    path?: string;
    requestId: string;
}
</pre>
One of the problems is that WebIDL is a well used notation in W3C specs, so we would need to have a good reason to replace it with something very similar.

wonsuk73:
@acrofts84 Let's discuss in the F2F. See you soon!

acrofts84:
Discussed at F2F. See #201

PR #209, opened 2017-05-25
Use Tables Instead of WebIDL

acrofts84:
Introduce tables instead of WebIDL for vssRequest, vssSuccessResponse and vssErrorResponse. 

This is for a demo before applying to the remaining interfaces.

@wonsuk73, what do you think?

### DIFF
--- a/vehicle_data/vehicle_information_service.html
+++ b/vehicle_data/vehicle_information_service.html
@@ -658,7 +658,7 @@ function subscriptionHandler(msg){
 	  </tr>
 	  <tr>
 	    <td> <dfn>action</dfn></td>
-	    <td> <a>Action</a></td>
+	    <td> <a href="#action">Action</a></td>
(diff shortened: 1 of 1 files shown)


wonsuk73:
@acrofts84 Thanks for the good proposal!
I agreed with the most of it! By the way there are few wrong links like vss link of vssSuccessResponse, requestId and error links of vssErrorResponse. Could you check that?^^

Still there are stuffs should be changed as you proposed like authorize, get, set, subscribe, unsubscribe, unsubscribeAll. I will take care of the last three items from subscribe. Is it good to you? :-)

acrofts84:
Hi Wonsuk. That sounds good, thanks :).  I will fix the links and get working on authorize, get and set.

wonsuk73:
@acrofts84 Thanks! I would like to merge your revised PR.

Issue #214, opened 2017-06-22
Duplication of host,protocol,port in VISClient and VISClientOptions


anawhj:
I found something strange while I read the VIAS specification carefully.

In VISClient Interface, 'host', 'protocol', 'port' attributes are defined twice in VISClient and VIS ClientOptions. If the attributes should be used in only creating an instance, the definitions in VISClient should be eliminated. If not, the definition of attributes in VISClient should be specified in the VISClient scope independent of VISClientOptions.

If I miss something about my understanding, please let me know the correction. I guess @aShinjiroUrata could handle this issue :)

aShinjiroUrata:
Hi @anawhj san,
Thanks for the comment!

> In VISClient Interface, 'host', 'protocol', 'port' attributes are defined twice in VISClient and VIS ClientOptions. 

The 'host', 'protocol', 'port' in `VISClientOptions` are for setting values at instance creation.
Those in `VISClient` are just for read the vaues (so, defined as `readonly`).

The reason 'host', 'protocol', 'port' are added in VISClient is, it would be inconvenient if those valuse
 are not available to read after the instance is created. 

What do you think?

anawhj:
Hi Urata san,

I'm Hyojin Song working in LG Electronics, and I'm simply looking into a couple of specs in Automotive WG. I could well understand your explanation in which the attributes are used for getting the value as `readonly`.

In that case, I would suggest that the attributes are defined in the scope of `VISClient`, and no need to define `VISClientOptions `interface. In addition, the constructor's parameters for `VISClient` would be mandatory, not optional depending on the usage.

You can see an example to review this issue in the following link.
https://www.w3.org/TR/dom/#interface-event

aShinjiroUrata:
Hyojin-san,

Thanks, I understand your point.

(1)style
With the current spec's style(style A), the code is like
```
var client = new VISClient( { 'host': 'wwwivi', 'protocol': 'ws', 'port':80 } );
```
and with your suggesting style (style B),
```
var client = new VISClient( 'wwwini', 'ws', 80);
```
I think both styles are found in other W3C specs.
style A example:
https://www.w3.org/TR/webrtc/#interface-definition (WebRTC  RTCPeerConnection)
style B example (as you wrote):
https://www.w3.org/TR/dom/#interface-event (DOM4 Event interface)

When there are many arguments, I suppose style A is tend to be selected.
In our case, I think, both style would work, and which means
,personally, I don't find a strong need to change to style B.
What do you think?

[...]

PR #215, opened 2017-06-22
[VIAS] fix warnings and errors from respec

anawhj:
To briefly understand the changes, the following list would be helpful.

- Change the spec draft version from `WD` to `ED` (`ED` is correct in the working space and `WD` is the result of publication in TR space)
- Modify some indentations (w/ eliminating tap indentations)
- Insert several `<def></def>` tag pairs and `data-dfn-for`, `data-link-for` attributes in the containing elements to specify the place of definitions for attributes, interfaces, and methods
- As a result, "51 warnings and 1 error" to "10 warnings and 0 error" (produced by Respec)

c.f. The remaining warnings from Respec could be fixed after additional descriptions for the following interfaces and its internal attributes: VISClientOptions([issue 214](https://github.com/w3c/automotive/issues/214)), VISSubscribeFilters, VISSubscribeRange

### DIFF
--- a/vehicle_information_api/vehicle_information_api_specification.html
+++ b/vehicle_information_api/vehicle_information_api_specification.html
@@ -6,7 +6,7 @@
     <script src='https://www.w3.org/Tools/respec/respec-w3c-common' class='remove'></script>
     <script class="remove">
       var respecConfig = {
-          specStatus: "WD",
+          specStatus: "ED",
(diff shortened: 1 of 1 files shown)


aShinjiroUrata:
Hi Hyojin san,

Thanks for the fix!
It looks good and merged it.

Much appreciated!

PR #216, opened 2017-06-22
Replace WebIDL definitions with table for subscribe

wonsuk73:
Replace WebIDL definitions with table for subscribe


wonsuk73:
@acrofts84 I made this using your proposal which I agreed. So I would like to merge this. If you have any comment on this, let me know.

PR #221, opened 2017-07-07
Update VSS Request section to clarify behaviour

acrofts84:
As discussed in #132


wonsuk73:
@acrofts84 Thanks for clarifed sentence with enough consensus. Merged!

Issue #210, opened 2017-05-25
Change "getVSS" to "discover"


acrofts84:
In order to further abstract the data model from the access methods, we should rename the "getVSS" method/action to the "discover" method/action.
- This is familiar parlance within the IoT and will be more familiar to an implementer.
- This ensures that the spec can still be logically applied to new data models, including but not limited to the VSS.

aShinjiroUrata:
Hi @acrofts84 san

I think abstraction of 'getVSS' is good idea.

By the way, although I am not very familiar with IoT specs, this word "discover" is appropriate for this use case?
In w3c wot, "Discovery AP" is to discover IoT "things" which is like discovering devices.
http://w3c.github.io/wot/architecture/wot-architecture.html

If the word "discover" is natural in this use case, I'm OK to change.

PeterWinzell:
I agree "discover" is better

drkevg:
Although we like 'discover', Adam and I think Urata-san raises a good point and that using 'discover' could be confusing if we use it to discover metadata and e.g WoT use it to discover services. Would it be ok if we replaced getVSS with getMetadata instead (we also considered getSchema, but the method call returns an object with metadata rather than a schema definition, so getMetadata seems clearest)

peterMelco:
Ok, I then change my mind :) getMetaData is more clear, so I am fine with that.

acrofts84:
I also agree with getMetadata

aShinjiroUrata:
Excuse me for late reply.
I'm OK with the word 'getMetadata'.

I looked more at OIC Core spec and looks, word 'discover' is used not only for 'Endpoint discovery',
  used for 'Resouce discovery' as well.
https://openconnectivity.org/developer/specifications

So, I thought something like 'discoverResouce' could be another candidate
 but it is not something like commonly used word.

So, in our case, I suppose 'getMetadata' matches well.

rstreif:
I agree that just 'discover' is too generic and somewhat overloaded. @acrofts84, looking at OIC is a good idea because of the interaction with them and making things consistent even across different applications makes a lot of sense. 'discoverResource' has my vote but I am ok with 'getMetadata' although it's a little too generic for my taste. :)

peterMelco:
So I will close this one since this is now changed in the spec to getMetaData

PR #237, opened 2017-09-18
Changed getVSS to getMetadata in VIAS like VISS #210

wonsuk73:
VISS spec had replaced getVSS with getMetadata due to further abstract the data model from the access methods. Which was discussed in issue #210 that is closed issue now. VIAS is need to be synchronized with VISS because it is abstract API of server interface spec for client application.

@aShinjiroUrata, could you review this?


aShinjiroUrata:
Thanks for the PR @wonsuk73 -san.
I have reviewed and merged.

Issue #242, opened 2017-11-25
Error Info(number, reason, message) to subscribe request with filter for branch


wonsuk73:
In current  [9. Server Side Filtering](https://w3c.github.io/automotive/vehicle_data/vehicle_information_service.html#server-side-filtering) part of VIS spec describe as below.

> Filters can only be applied for data which is represented by a primitive type, such as leaf nodes in the VSS tree, not entire branches. For example, a filter cannot be set on Signal.Drivetrain.InternalCombustionEngine.*, however it can be set on Signal.Drivetrain.InternalCombustionEngine.RPM.

By the way I couldn't find appropriate error number, reason and message in the [9.2 Errors](https://w3c.github.io/automotive/vehicle_data/vehicle_information_service.html#errors) part. If it is true it should be added.

wonsuk73:
@acrofts84 @drkevg @aShinjiroUrata Does it make sence to you guys?

acrofts84:
Hi Wonsuk. 

I think the logical response to this would be a 400 Bad Request, but I agree that the error message "The server is unable to fulfil the client request because the request is malformed." doesn't provide a satisfactory response. 

How about introducing 400 Bad Request "Filter requested on non-primitive type"?

wonsuk73:
@acrofts84 That's good to me! But to be clear, error code is "400(Bad Request)" and message is "Filter requested on non-primitive type".So Error Reason need to be added. Right?

acrofts84:
Good point. I think the reason has to be a bit more concise, so I think the best approach is:

Code: 400 (Bad Request) 
Reason: filter_invalid
Message: Filter requested on non-primitive type

wonsuk73:
@acrofts84 Thanks!

PR #248, opened 2017-11-30
Adding error info regarding wrong target of filter  #242

wonsuk73:
Filter cannot be applied to  the branch so it needs to define additional error info for the violated case.
- error code: 400(bad request)
- error reason: bad_filter_requested
- error message: Filter requested on non-primitive type


acrofts84:
Thanks Wonsuk :)

wonsuk73:
@acrofts84 I think "filter_invalid" is better string rather than "bad_filter_requested" for error reason as you wrote. so I would like to change that. :-)

acrofts84:
No problem. Thanks!

PR #259, opened 2018-02-14
fixing typos

dret:


wonsuk73:
@dret, Thanks for contribution! Sorry for late response.

PR #260, opened 2018-02-14
Consolidated the use of American and British English

TalAter:
Changed "authorised" to "authorized" as this spelling was only used in 2 places throughout the document, while the common spelling (including in action names, and tokens) seems to favour (or rather favor) the American English spelling.

Please let me know if:

1. I should continue to consolidate spelling throughout the document on either style
2. I shouldn't bother with trivialities like this
3. I should steer clear of topics like this which are too explosive to reach consensus on

### DIFF
--- a/vehicle_data/vehicle_information_service.html
+++ b/vehicle_data/vehicle_information_service.html
@@ -1063,8 +1063,8 @@ function subscriptionHandler(msg){
         <p>The Vehicle Signal Server (VSS) specification allows branches to be defined as either public or private. The VIS server will 
 	satisfy a request for metadata in a public branch regardless of the access control permissions associated with the client 
 	making the request. This means that a user could see that a particular signal in a public branch is available but they may not 
-	be currently authorised to Get, Set, or Subscribe to that signal. If a branch is defined to be 'private', only suitably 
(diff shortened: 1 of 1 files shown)


PeterWinzell:
I rember us actually taking the decision to use British , since two of our editors were English

TalAter:
I have no personal preference either way, but it seems to be using the American English spelling much more often - including in the code.

automotive_vehicle_information_service_html_at_be26851c75f89d93e5ba84853d7f5b42aa696cab_ _talater_automotive

gunnarx:
I vote for 1. It's quality improvement, so yes it ought to be done.  
The document editor(s) however need to agree and communicate about a policy.

drkevg (approved):
Hi Tal, thanks for spotting these.  We have 55 examples with 'Z' and these two with 'S' are anomalies that we had missed. So agree that they should be changed for consistency. 
Re. which is correct, many native British writers use either authorise or authorize, whereas its our understanding that only the form with 'Z' is valid in US, so it seemed like using 'Z' would achieve widest consensus.

drkevg:
Agreed that should change two instances of 'authorise' to 'authorize' for consistency

PR #264, opened 2018-04-16
Implementation report

aShinjiroUrata:
Pull request of "implementation_report.html".

Could you review?
(Especially, @peterMelco since creater of document, and @wonsuk73 since name is mentioned.)


wonsuk73:
I have merged to see the shape of implementation report! I believe you guys ;)

PR #271, opened 2018-08-07
Update services-explainer.md

PeterWinzell:
Changing the intent text of where we applications that uses this standard can reside.


wonsuk73:
@tguild Could you review this? I guess your review is mandatory because you made original texts. As a reference I am okay with it.

peterMelco:
Ok

tguild (approved):
Agree although also wanted to be clear automobiles generally should not be directly addressable on the web. Cloud based use cases would need additional consideration since it will not be live signals but recorded/delayed and we do not have mechanism for time based call back.

Thank you for the contribution.

PR #278, opened 2018-11-22
Generation two

UlfBj:
Html spec files created as decided on WG meeting 181119.


wonsuk73:
@UlfBj I think you are trying to create interface directory in automotive repo and put two files to there. right? I guess files are okay but I am not sure the directory name, interface. Do you have any idea for better name of directory to match with the work? I don't have strong intention but if we have better directory name to be matched with those files, that would be great to look for the files from contributors :)

tguild:
we can change directory name later if needed

PR #288, opened 2019-01-21
rephrase data model description, describe tree elements and add adressing chapter

PatrickCQ:


UlfBj:
The clause describing the general tree node types mentions signal/attribute node types. the signal type does not exist anymore, it is broken up into sensor/actuator/stream nodes types. The signal terminology could be used as an umbrella name for all of these, but I think that should be avoided as it is associated with the previous data model generation in VSS.  This description does not mention the new "resource" extension, i. e. the rbranch and element node types.  
The table has a line for "signal" type which I think should be replaced by the above mentioned types.

PatrickCQ:
"signal" is only there for now as it is described in the graphics. Will be replaced in the new types.

PR #289, opened 2019-01-25
Wildcard usage over HTTP chapter added.

UlfBj:


spec/Gen2_Core.html
```diff
@@ -99,6 +99,34 @@ <h2>Examples</h2>
...
         </section>
+        <section data-dfn-for="Foo">
+          <h2>Wildcard usage over HTTP</h2>
+          <p>The <a href='#rfc3986'>[RFC 3986 "Uniform Resource Identifier (URI): Generic Syntax"]</a> declares the asterix (*) as a reserved character, which then must be percent encoded if to be present in the path component of a URI. For readability and convenience reasons an alternative HTTPURL encoding is introduced here, which is to replace the asterix by the in RFC 3986 unreserved character tilde  (~).<br>
+          <blockquote>
+            The wildcard character asterix (*) used in the VSS path shall in HTTP URLs either be percent encoded as (%2A), or represented by the tilde character (~).<br>
+          </blockquote>
+          <blockquote>
+            The use of wildcards must be supported when a client connects to the server via the Websocket protocol.<br>
+            The use of wildcards is optionally supported when a client connects to the server via the HTTP protocol. If not supported by the server, the error code XXX shall be returned.<br>
```
gunnarx: XXX code:  I assume it must be a 404 error code - the requested URL does not exist?  This is not so informative to explain that it is an unsupported feature, but it seems formally the correct one?

PR #292, opened 2019-02-05
Interface

PatrickCQ:
reworked the interface section


UlfBj:
Row 191 in the diff file mentions a filter parameter for the read method:
"<li><a>filter</a> selection criterias for specific values under the given <a>path</a></li>"
Which can also be found as a parameter in other methods. 
This concept is unclear to me, the terminology and concept is understandable for the subscription method,as it is used in VISS, but for other methods the term query seems more natural to me. And if so, I do not see the query to be an explicit parameter, but an extension of the path parameter. 
Please enlighten me on this.

PatrickCQ:
> Row 191 in the diff file mentions a filter parameter for the read method:
> "
> 
> * filter selection criterias for specific values under the given path
> "
> Which can also be found as a parameter in other methods.
> This concept is unclear to me, the terminology and concept is understandable for the subscription method,as it is used in VISS, but for other methods the term query seems more natural to me. And if so, I do not see the query to be an explicit parameter, but an extension of the path parameter.
> Please enlighten me on this.

For me the query describes that everything that is asked of another party. So this includes the path and directive (e.g. get a spec for that path or get a value in that path) and the rules to filter something.
I wanted to make clear that the filter is just there to reduce the sub-tree / the result but not to describe it (that would be the full query). 

That seems to be a great topic to be discussed tonight.

PatrickCQ:

PR #294, opened 2019-02-06
Addressing

PatrickCQ:
write addressing chapter and adapt examples


PatrickCQ:

PR #295, opened 2019-02-08
Filter - from query to filter

PatrickCQ:
describe filtering on tree


UlfBj:
It seems the filter model described here is abandoning the wildcard mechanism. It may be ok, but then wildcard expressions should not be used in the description of the model. 
Are the query rules expressed here, and its syntax, used elsewhere before?
If not, we need to scrutinize it very carefully before bringing it aboard. 
There is an ongoing discussion at VSS to abandon wildcards, together with introducing changes to the model and tools. Maybe we should wait for that to settle before concluding here?

PeterWinzell (commented):
1) "_Most times it is not the best way to get a whole part of the data representation from the service when looking for a specific set with a specific value"_. This section describes how to build the queries to search for specific parts of the data representation.

**The first sentence is a bit long and i actually think it does not add any real value. I would skip and just use "This section..."**

2)
 ...is an expression that describes a characteristic of nodes that shall be returned by a <a>query</a>.
			<a>Rules</a> are applied to a leaf and **descibe** a desired attribute of a node...

typo: should be describe.

spec/Gen2_Core.html
```diff
@@ -298,51 +298,175 @@ <h2>Service discovery</h2>
...
+	  <ul>
+			<li>path: describing the part of the tree that shall be filtered in</li>
+			<li>rules: these have to match in the nodes for the node to be part of the <a>content</a> response.</li>
+	  </ul>
+		</p>
+		<div class="example">
+		<h3>Example</h3>
+			<p>
+			All open doors shall be given. The tree looks as follows:
+			<blockquote>
```
danielwilms, line 318: will be outdated in V2

spec/Gen2_Core.html
```diff
@@ -298,51 +298,175 @@ <h2>Service discovery</h2>
...
+							</li></ul>
+						</li></ul>
+					</li></ul>
+				</li></ul>
+			</blockquote>
+			</p>
+			<p>
+			The <a>path</a> that shall be inspected is "/vehicle/cabin". <br>
+			The <a>rule</a> is that "/vehicle/cabin/*/*/Door/isLocked" shall be "true".<br>
+			</p>
```
danielwilms, line 362: Consider https://github.com/GENIVI/vehicle_signal_specification/issues/81 as input. would already take that one as base.

PR #296, opened 2019-02-08
Plantuml

PatrickCQ:
replace tree description by plantuml.
Changing graphics is hard if the base file is not available. With plantuml it is just a change of lines.


gunnarx:
* +1 for PlantUML, generally.
* But:  I know W3C specs are written directly in a HTML document, but is it OK to add any Javascript libraries we want to the spec?
* I would clean up the git history a little, at least squash the last two commits together.  "Replace" means both add and remove.
* Reviewing the result is a bit challenging.  I don't know a way to view GitHub pages for a PR (or alternative branch).  I did a quick Google check but did not find it.  Both the HTML file and the Javascripts are needed, so I checked out the whole PR from git.  It looks alright, the "Component" look for each box feels like a bit of a compromise - if we could have plain boxes it might be even better but I don't know how?

One option for easier review might be that you merge this to your master branch, then we may be able to view it on PatrickCQ.github.io (if you enable it)?

PatrickCQ:
> * Reviewing the result is a bit challenging.  I don't know a way to view GitHub pages for a PR (or alternative branch).  

If just viewing the files of a PR it should be possible to review the changes (https://github.com/w3c/automotive/pull/296/files)

PatrickCQ:
> * I would clean up the git history a little, at least squash the last two commits together.  "Replace" means both add and remove.

I have left in commits on purpose to give hints to the evolution of the document.

PatrickCQ:
> * But:  I know W3C specs are written directly in a HTML document, but is it OK to add any Javascript libraries we want to the spec?

@tguild could you check if this is ok?

gunnarx:
> If just viewing the files of a PR it should be possible to review the changes (https://github.com/w3c/automotive/pull/296/files)

Well... I meant that it is not possible to see what the plantUML rendering looks like there.  (I thought that was clear).  It's possible to see that it in other ways of course, some of which I described.  I just intended to make some proposals to make the review process simpler for the others.

tguild:
> > * But:  I know W3C specs are written directly in a HTML document, but is it OK to add any Javascript libraries we want to the spec?
> 
> @tguild could you check if this is ok?

We can include JS on specs. Preference would be for us to serve a copy of the JS instead of a third party to avoid privacy/tracking concerns, security etc.

PatrickCQ:
> > If just viewing the files of a PR it should be possible to review the changes (https://github.com/w3c/automotive/pull/296/files)
> 
> Well... I meant that it is not possible to see what the plantUML rendering looks like there. (I thought that was clear). It's possible to see that it in other ways of course, some of which I described. I just intended to make some proposals to make the review process simpler for the others.

good idea. Let us use raw githack for that:
https://rawcdn.githack.com/PatrickCQ/w3c-automotive/plantuml/spec/Gen2_Core.html

gunnarx:
Nice one.  That magically works with the javascript references too (because they are relative URIs, and so it becomes a valid URL, also fetched from raw.githack).   I suppose that would be happen also for raw.github, if only the original file was went with the right Content-Type...

PatrickCQ:

PR #297, opened 2019-02-08
bring Data model to current VSS state

PatrickCQ:
As we are trying to make the changes from VSS clear to everybody we have to start from the current VSS state.
This should reflect it.


PatrickCQ:

PR #298, opened 2019-02-15
Data model gen2

PatrickCQ:
This is a description of the data model as I envision it after going through the discussions of the last year, RSI and VISS.
Easy to read version: https://raw.githack.com/PatrickCQ/w3c-automotive/data-model-gen2/spec/Gen2_Core.html

### DIFF
--- a/spec/Gen2_Core.html
+++ b/spec/Gen2_Core.html
@@ -323,7 +323,7 @@
           A <a>remote branch</a> is <a>branch</a> that MUST point to another <a>service</a> that implements <a>branches</a> to represent data.
           The other <a>services</a> are called remotes. At least one <a>remote</a> MUST be defined. A <a>remote</a> must contain a <a>schema</a> and a <a>authority</a>.
           
-          It MAY point to a sub-part of the <a>tree</a> by setting a <a>path</a>. If no <a>path</a> is given <code>/</code> is assumed.
+          It MAY point to a sub-<a>branch</a> by setting a <a>path</a>. If no <a>path</a> is given <code>/</code> is assumed.
           
           </p>
             <figure id="remote-branch-example" class="example">
--- a/spec/Gen2_Core.html
+++ b/spec/Gen2_Core.html
@@ -62,6 +62,11 @@
       Leafs in the tree are <a>values</a>. 
       The path to a leaf is built by <a>branches</a>.
       </p>
+      <p>
+        A <dfn>schema</dfn> for a <a>branch</a> describes the structure of the <a>branch</a> itself 
+        and the available child <a>branches</a> 
+        and the optional and mandatory <a>values</a>.
+      </p>
       <figure id="tree-example" class="example">
         <img uml='
         @startuml 
@@ -155,7 +160,7 @@
           </p>
         </section>
       </section>
-  
+      
       <section data-dfn-for="address">
         <h2>Addressing</h2>
         <p>Addressing of elements is done using URIs as defined in [[RFC3986]].</p>
--- a/spec/Gen2_Core.html
+++ b/spec/Gen2_Core.html
@@ -51,7 +51,7 @@
         The Gen2 messaging layer uses a RESTful design for all methods exchanged via the interface (<a href="#interface"></a>).
       </p>
       <p>
-        Then Gen2 data model (<a href="#data-model"></a>) is the same through all transport protocols.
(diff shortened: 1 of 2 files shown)


gunnarx:
About the comment in the document:

> " is this the correct link?"

Yes, I would agree that the link you are using:
https://github.com/GENIVI/vehicle_signal_specification/blob/master/README.md
is fine if you want the README specifically, and as we can see it will always be showing the latest version of it existing on the master branch.

Alternatively I think https://github.com/GENIVI/vehicle_signal_specification might be more general since it points to the actual repository with all the files, and the README is still rendered below it.  (That's a style choice.  Point to the actual VSS database, or prefer a document.)

If the repository moves or is renamed, I expect that GitHub might be smarter about getting people to the right place using the link to the "base" of the repository only, but that's just a guess.  Also, we can note that such a "base" link does not specify the branch.  It will always show the branch that we have configured as the default branch in GitHub.  (This is master now, but  could be modified in GitHub to be something other than master).  That feature could be an advantage (or not) depending on the scenario.

In summary I think the introductory text that the link appears in, is not intended to lock down one specific commit/version of the database (it is under development), so going with the shorter and more general link should be the best choice, in my opinion.  Both of them are OK though.

UlfBj:
Regarding the Data model and following clauses Tree elements, Addressing, Value Data types, etc
Shall this be described here?
We point to VSS, so I think we shall refer to the data model description there, and not have one here also. It will only lead to ambiguity, when they start to differ (which seems to be the case already). If we are not satisfied with the VSS data model, we should submit VSS PRs to request changes.

PatrickCQ:
> more general link

linking to the more general part.

PatrickCQ:
> Regarding the Data model and following clauses Tree elements, Addressing, Value Data types, etc
> Shall this be described here?
> We point to VSS, so I think we shall refer to the data model description there, and not have one here also. It will only lead to ambiguity, when they start to differ (which seems to be the case already). If we are not satisfied with the VSS data model, we should submit VSS PRs to request changes.

It is my understanding that we started from VSS and are moving to something different that is defined by the W3C automotive working group. 
The changes suggested in this PR are influencing the style of the VSS description in its base without removing its functionality.

UlfBj:
I think we should be very careful before going in that direction. Personally, I do not share that view.

spec/Gen2_Core.html
```diff
@@ -67,8 +67,7 @@ <h2>Data model</h2>
       It is represented by a directed acyclic graph.
       </p>
       <p>
-      It is derived from <a href="https://github.com/GENIVI/vehicle_signal_specification/blob/master/README.md">GENIVI's Vehicle Signal Specification</a> (VSS).
-      <span class="issue">is this the correct link?</span>
+      It is influenced by <a href="https://github.com/GENIVI/vehicle_signal_specification">GENIVI's Vehicle Signal Specification</a> (VSS).
```
gunnarx: In the next meeting, let's talk about the rationale for introducing a softer language here.  It would actually seem more accurate to me to change the wording in the other direction since the database of signals is basically **defined** by the VSS, currently.

Over time, a new database may evolve, but surely it would make sense for the VSS to follow that database evolution.  

If you feel this data format is only influenced then it must be something different?  If you can let us know in what way is VSS not providing the right format, there is the potential to change it.

gunnarx (changes requested):

gunnarx:
> VSS and are moving to something different

I share Ulf's way of describing it, that we are actually adjusting VSS to match whatever that "different" thing is.   Consider it from VSS perspective - I don't yet see a strong need for the repo at GENIVI, or any copy anywhere else, to keep some old meaning for the term "VSS" that is no longer connected to Gen 2.   At least so far, the idea (or need) has not come up for these to diverge in different directions?   

Also, as far as I can tell, the group already agreed to progress a "version 2.0" of the VSS repository - wasn't that the purpose of all the discussion we had about what should be the master branch, and what should be the maintenance branch?

PeterWinzell:
So, based on todays discussion I find the question at which we stand as group to be wether we should have a generic description of a data model for accessing vehicle data and vehicle sensor data, or if we should continue using VSS as the description of the data model. 
a) merge proposal , generic description of the data model
b) data model description === VSS 

I think that we should continue with b , since this points to an existing data model that have been used and is being used (vss 1.0).

PR #299, opened 2019-02-17
meta-data/headers

PatrickCQ:
Description of meta-data/headers that are known when methods are being executed.
E.g for authentication tokens, schema variant description . . .

Easy read version: https://raw.githack.com/PatrickCQ/w3c-automotive/method-extension/spec/Gen2_Core.html

### DIFF
--- a/spec/Gen2_Core.html
+++ b/spec/Gen2_Core.html
@@ -343,10 +343,6 @@
         Authentication: <a href="http://www.w3.org/2018/04/19-auto-minutes.html#item03">http://www.w3.org/2018/04/19-auto-minutes.html#item03</a><br>
         Volvo assesment: <a href="https://lists.w3.org/Archives/Public/public-automotive/2018Aug/0002.html">https://lists.w3.org/Archives/Public/public-automotive/2018Aug/0002.html</a>, <a href="https://www.w3.org/2018/09/25-auto-minutes.html#item01">https://www.w3.org/2018/09/25-auto-minutes.html#item01</a>
       </div>
-      <section data-dfn-for="Authorize">
-        <h2>Authorize</h2>
-        <p>To enable access to signals and data attributes that are under access control, the client MAY optionally pass a message with an authorize action to the server. If the server is able to satisfy the request it SHALL return a Response message. If an error occurs the server SHALL return an ErrorResponse message. The key data components of these messagess are defined below:</p>
-      </section>
     </section>
 
     <section data-dfn-for="interface">
--- a/spec/Gen2_Core.html
+++ b/spec/Gen2_Core.html
@@ -678,6 +678,7 @@
             The client MUST set the type of the token with the <a>AuthType</a> key/value pair if <a>AuthToken</a> is set.
             This concept MAY be used for access control.
             A rejected authorization token MUST result in an <a>error</a> response.
+            If the execution of a method on a service needs authorization the service MUST use the authorization key/value pair.
             <table class="def">
               <tr><th>key</th>    <td><dfn>AuthToken</dfn></td></tr>
               <tr><th>value</th>  <td><code>^\w{1,1024}$</code></td></tr>
--- a/spec/Gen2_Core.html
+++ b/spec/Gen2_Core.html
@@ -538,11 +538,53 @@
         </section>
         <section data-dfn-for="error">
         <h2><dfn>error</dfn></h2>
-        <ul>  
-          <li>Error: Information describing the error reason.</li>
-          <li>Timestamp: A timestamp associated with the failed request.</li>
-        </ul>
+          <p>
+          An <a>error</a> response MUST include the formal error type.
+          An <a>error</a> response MAY include a textual form of the error type. 
+          An <a>error</a> response MUST include an error value if obligatory for the error type.
+          An <a>error</a> response MUST include a timestamp.
+          <table class="def">
+            <tr>
+              <th>field</th><th>obligation</th><th>description</th><th>format</th><th>example</th></tr>
+            <tr>
+              <th>error type</th>
+              <td>mandatory</td>
+              <td>short handle describing the error.</td>
+              <td>protocol specifc</td>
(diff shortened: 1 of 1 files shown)


spec/Gen2_Core.html
```diff
@@ -343,16 +351,20 @@ <h2>Security</h2>
...
-      <section data-dfn-for="Authorize">
-        <h2>Authorize</h2>
-        <p>To enable access to signals and data attributes that are under access control, the client MAY optionally pass a message with an authorize action to the server. If the server is able to satisfy the request it SHALL return a Response message. If an error occurs the server SHALL return an ErrorResponse message. The key data components of these messagess are defined below:</p>
-      </section>
     </section>
 
     <section data-dfn-for="interface">
       <h2>Interface</h2>
+      <p>
+        This chapter describes the different methods and its arguments that regulate the communication between a <a>service</a> and a <a>client</a>.
```
ghost: Not sure about the word "regulation" here. I would have used "govern" maybe ? but I might just be nitpicking.

spec/Gen2_Core.html
```diff
@@ -528,11 +540,176 @@ <h2><dfn>contract</dfn></h2>
...
+              </tr>
+            </table>
+            <div class="issue" title="error type examples">Would examples for the error types be helpful?</div>
+            <div class="issue" title="error type location">Where best to describe error types?</div>
+          </section>
+        </section>
+        
+      </section>
+      <section data-dfn-for="headers">
+        <h3>Meta-Data</h3>
```
ghost, line 661: Meta data, out of band, etc. Is the purpose of this section to establish minimum requirements for illegible transports or to provide  a fallback for a transport that might not support  natively token based authentication ?

spec/Gen2_Core.html
```diff
@@ -528,11 +540,176 @@ <h2><dfn>contract</dfn></h2>
...
+        <p>
+          For the communication between service and client several key/value pairs can be given to describe the environment in what the methods (<a href="#methods"></a>) are called.
+          The string describing the key MUST be the same structure as <a>name</a>.
+        </p>
+        
+        <section data-dfn-for="authentication">
+          <h4><dfn>authentication</dfn></h4>
+          <p>Purpose: <a>authentication</a> makes it possible for a client to authenticate itself towards a service.</p>
+          <p>
+            The <a>authentication</a> key/value pairs consists out of a <a>AuthType</a> and a <a>AuthToken</a> key/value pair.
```
ghost: What are the requirement on the Server side?

spec/Gen2_Core.html
```diff
@@ -528,11 +540,176 @@ <h2><dfn>contract</dfn></h2>
...
+            The client MAY use the <a>authentication</a> key/value pairs to transfer an authentication token to the service.
+            The client MUST use either both key/value pairs or none of the two.
+            The service MAY use the token transfered by the <a>AuthToken</a> key/value pair to authenticate the client.
+            The client MUST set the type of the token with the <a>AuthType</a> key/value pair if <a>AuthToken</a> is set.
+            This concept MAY be used for access control.
+            A rejected authentication token MUST result in an <a>error</a> response.
```
ghost: Please specific which type of error.

spec/Gen2_Core.html
```diff
@@ -528,11 +540,176 @@ <h2><dfn>contract</dfn></h2>
...
+              <tr><th>value</th>  <td><a>name</a> compliant</td></tr>
+            </table>
+          </p>
+        </section>
+        
+        <section id="variant-selection">
+          <h4>variant selection</h4>
+          <p>
+            Purpose: 
+            A client can express which schema variants it can support when executing a method.
```
ghost, line 698: Schema of the payload?

UlfBj:
Gen2_Core.html: Row 670 and onwards
Authentication or authorization?
The token concept is mainly a way to request authorization than provide authentication, as I understand it.

spec/Gen2_Core.html
PatrickCQ: updated language

PatrickCQ:
> Gen2_Core.html: Row 670 and onwards
> Authentication or authorization?
> The token concept is mainly a way to request authorization than provide authentication, as I understand it.

We discussed if the generation of tokens should be within the scope of Gen2.
It was my understanding that this is not the case (https://www.w3.org/2018/09/25-auto-minutes.html#item02).

UlfBj:
I agree with that the generation of tokens should be out of scope. What I am trying to say is that the token usage is more intended for an authorization purpose than for an authentication purpose. The authentication is primarily handled as part of the token generation procedure. 
So I believe we should thus rather speak about authorization than authentication in this context.

spec/Gen2_Core.html
PatrickCQ: payload is described as to be "schema" compliant
https://github.com/w3c/automotive/blob/2a5b0e319e0e26206a2df6dc42530aa180d66913/spec/Gen2_Core.html#L707

spec/Gen2_Core.html
PatrickCQ: That has to be done for all errors.
https://github.com/w3c/automotive/blob/2a5b0e319e0e26206a2df6dc42530aa180d66913/spec/Gen2_Core.html#L367

PatrickCQ:
> So I believe we should thus rather speak about authorization than authentication in this context.

Now I get it. Yes.
https://github.com/w3c/automotive/blob/73f5c27447d6ba53227497cfa52a1f102eb3a8c5/spec/Gen2_Core.html#L670-L691

spec/Gen2_Core.html
PatrickCQ: Is this style of authorization to be used if authorization is necessary with the service?

PatrickCQ: added language to spec

PatrickCQ:
Group ok with merging.
https://www.w3.org/2019/02/19-auto-minutes

PR #302, opened 2019-04-03
Data types

PatrickCQ:
The intention of this PR is to reduce the amount of data-types to be implemented in a transport protocol and to be handled by all implementations manipulating data.

### DIFF
--- a/spec/Gen2_Core.html
+++ b/spec/Gen2_Core.html
@@ -275,8 +275,46 @@
         
         <section data-dfn-for="String">
           <h3><dfn>String</dfn></h3>
-          <p>Type: character string</p>
+          <p>
+          <a>String</a> is used to represent sequence of characters.
+          To allow for memory allocation restrictions MUST be given.
+          Restrictions MUST include a maximum string length.
+          Restrictions MAY include a character set.
+          If no character set is given unicode MUST be assumed.
+          Restrictions MAY include a regular expression.
+          <span class="issue">which regex style?</span>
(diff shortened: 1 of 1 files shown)


PeterWinzell (commented):
1)
Resolution MUST . ... 
must what ?  some part missing here.


2)
_of not nearer specified data_ ... **is that not what a byte array is ?**

3)
_also known as a blob is an byte array_ : I think it should be **a** byte array

ghost:
1. Resolution or Precision? In case of floating point numbers I would favor the later. For floating point number the encoding should be specified.
2. Note sure about the naming of "ByteBuffer", ByteArray of Sequence would be more appropriate I think.

Should we express that the intented target of min/max/resolution is the function that will make use of the data, not the serializer (i.e. it takes 3 bytes to express 255 in json but the data itself is really only one byte)

spec/Gen2_Core.html
```diff
@@ -190,79 +190,142 @@ <h2>Addressing</h2>
...
+              <td>-2147483648</td>
+              <td>2147483647</td>
+              <td></td></tr>
+            <tr><th><dfn>UInt64</dfn></th>
+              <td>unsigned 64-bit integer</td>
+              <td>0</td>
+              <td>2^64-1</td>
+              <td></td></tr>
+            <tr><th><dfn>Int64</dfn></th>
+              <td>unsigned 32-bit integer</td>
```
ghost: unsigned 32 bit?

PR #304, opened 2019-04-04
Simple tree elements

PatrickCQ:
This PR describes a simplification of the available elements which have to be implemented for all transport protocols.

### DIFF
--- a/spec/Gen2_Core.html
+++ b/spec/Gen2_Core.html
@@ -276,6 +276,92 @@
           <div class="issue">to be described more</div>
         </section>		
       </section>
+
+      <section data-dfn-for="branch-types">
+        <h2>Default Branch Types</h2>
+        <div class="issue" title="branch types">
+          This is an example on how branches could be used. The idea is very much the same as with the predefined restrictions for data-types. 
+        </div>
+        <p>
+        A set of default <a>branch</a> types is defined. 
+        These types help to have a common understanding how a branch can be understood 
+        and allows for easy reuse of code throught several schemas.
+        These types MUST be implemented.
+        </p>
+        <section data-dfn-for="remote-branch">
+          <h3><dfn>remote branch</dfn></h3>
+          <p>
+          A <a>remote branch</a> is <a>branch</a> that MUST point to another <a>service</a> that implements <a>branches</a> to represent data.
+          The other <a>services</a> are called remotes. At least one <a>remote</a> MUST be defined. A <a>remote</a> must contain a <a>schema</a> and a <a>authority</a>.
+          
+          It MAY point to a sub-<a>branch</a> by setting a <a>path</a>. If no <a>path</a> is given <code>/</code> is assumed.
+          
+          </p>
+            <figure id="remote-branch-example" class="example">
+          <img uml='
+          @startuml 
+            skinparam usecase {
+              FontSize 16
+              BackgroundColor CornflowerBlue
+              FontColor Black
+              BorderColor Black
+            }
+
+            skinparam component {
+              FontSize 16
+              BorderColor Black					
+              BackgroundColor MediumSeaGreen
+              FontColor Black
+            }
+
+            [Root] -- [Media]
+            [Root] -- [Car]
+
+            note left of Media: remote branch
+            [Media] -- [Collections]
+            [Collections] -- [Remotes]
+            [Collections] -- (path\n"/collections")
+            [Collections] .. [remote]
+            
+            note left of Remotes: list of remotes
+            
+            [Remotes] -- (Scheme\n"http")
+            [Remotes] -- (Authority\n"127.0.0.1:1337")
+            
+            
+            package "Remote http://127.0.0.1:1337/Collections" as remote {
+              [30321564123] -- (name)
+              [30321564123] -- (rating)
+            }
+            
+            note bottom of remote: Remote representation of a part of a tree
+            
+            [Car] -- [Engine]
+            [Car] -- [...]
+            
+            [Engine] -- (RPM)
+            [Engine] -- (Speed)
+          @enduml
+          '>
+              <figcaption>Example of a remote branch</figcaption>
+            </figure>
+      
+        </section>
+        <section data-dfn-for="remote">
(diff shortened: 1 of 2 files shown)


spec/Gen2_Core.html
```diff
@@ -54,132 +54,132 @@ <h2>Terminology</h2>
...
+        @enduml
+        '>
         <figcaption>Diagram showing an example tree.</figcaption>
       </figure>
 	    <section data-dfn-for="tree-elements">
         <h2>Tree elements</h2>
-        <section data-dfn-for="rbranch">
-          <h3><dfn>rbranch</dfn></h3>
+        <p>
+        Tree elements have <dfn>names</dfn>.
```
danielwilms, line 142: Talking about VSS the name is the path to the node. Is it meant like that, or is the name another concept?

spec/Gen2_Core.html
```diff
@@ -54,132 +54,132 @@ <h2>Terminology</h2>
...
+        A <a>name</a> MUST be a string with at least 3 characters.
+        A <a>name</a> SHALL NOT be longer than 32 characters.
```
danielwilms, line 144: Can several elements have the same name?

spec/Gen2_Core.html
```diff
@@ -54,132 +54,132 @@ <h2>Terminology</h2>
...
+        This RegEx expresses <a>name</a>: <code>^[A-Z][A-Za-z0-9]{2,31}$</code>
+        <div class="issue" title="properties of name">The properties of name must be defined. This is just expressing a gut feeling.</div>
+        <section data-dfn-for="branch">
+          <h3><dfn>branch</dfn></h3>
           <p>
-          A node under an <a>rbranch</a> MUST be an <a>element</a>.
+            A <a>branch</a> MUST have a schema attached that describes it sub-nodes. 
+            It MUST be marked for every sub-<a>branch</a> if only one instance is allowed or multiple instances are allowed.
+            A <a>branch</a> MUST have a <a>name</a> that makes the <a>branches</a> convention/type understandable.
+            The <a>name</a> is set by the schema.
```
danielwilms, line 155: >  The <a>name</a> is set by the schema. 

Do you have an example?

UlfBj:
In the WG meeting april 9 the survey results clearly pointed to that the group wanted to go with VSS. This is not VSS, therefore I do not support this PR. 
If anything of this PR should become used in this specification, it should be pushed to the VSS repo, and not to this. 
Specifying parts of the data model here, and other parts in VSS, is not a good practice.

spec/Gen2_Core.html
```diff
@@ -54,132 +54,132 @@ <h2>Terminology</h2>
...
 		It is represented by a directed acyclic graph.
-		</p>
-		<p>
-		It is derived from <a href="https://github.com/GENIVI/vehicle_signal_specification/blob/master/README.md">GENIVI's Vehicle Signal Specification</a> (VSS).
-		<span class="issue">is this the correct link?</span>
+      </p>
+      <p>
+        It fully supports the <a href="https://github.com/GENIVI/vehicle_signal_specification">GENIVI's Vehicle Signal Specification</a> (VSS) 
+        which is recommended by the W3C [[vehicle-information-service]].
+        This description gives the template, the rough outlines, out of which parts an ontology can be build.
```
danielwilms, line 64: >  out of which parts an ontology can be build

what do you mean with that?

spec/Gen2_Core.html
```diff
@@ -54,132 +54,132 @@ <h2>Terminology</h2>
...
-          A node under an <a>rbranch</a> MUST be an <a>element</a>.
+            A <a>branch</a> MUST have a schema attached that describes it sub-nodes. 
+            It MUST be marked for every sub-<a>branch</a> if only one instance is allowed or multiple instances are allowed.
+            A <a>branch</a> MUST have a <a>name</a> that makes the <a>branches</a> convention/type understandable.
+            The <a>name</a> is set by the schema. 
+            <div class="issue" title="type of branch">not clear that this expresses the type of the branch.</div>
+          </p>
+          <p>
+            A <a>branch</a> MUST have an <dfn>id</dfn> that makes the identifiable and addressable.
+            An <a>id</a> MAY be a unique string within a parent <a>branch</a> with the same properties as <a>name</a>.
```
danielwilms, line 160: Understood from last meeting, that this can be generated from the VSS name. Could we add that as an example/recommendation?

spec/Gen2_Core.html
```diff
@@ -276,6 +276,92 @@ <h3><dfn>branch</dfn></h3>
           <div class="issue">to be described more</div>
         </section>		
       </section>
+
+      <section data-dfn-for="branch-types">
+        <h2>Default Branch Types</h2>
+        <div class="issue" title="branch types">
```
danielwilms, line 282: seems to be a duplication from VSS docs, which should be used as snapshot as far as I understood the voting

gunnarx:
I have resolved the conflicts also on this, but the content is the same as before so that's the decision that needs to be made first and there is not much point in pushing the rebased version, if it's not to be merged.

I do not think there have been new significant changes on the descriptions of the data model, in other words we do not have two opposing/conflicting texts.  In this case, rather some parts were not accepted with the view that the proposed text should not be added at all.

I guess this is one of those commits: https://github.com/w3c/automotive/pull/304/commits/3c18080b4defd6d277934cdb962de11a6dcc6b0f, and this https://github.com/w3c/automotive/pull/304/commits/6fb5d43f52d270dc3b98b057f8f486224cc190ec

Overall, I would propose that reviewers should try to specify which commits might be acceptable and which might not, so that not some useful part gets lost if a PR is rejected completely.

For example, might not this change be useful nonetheless?  https://github.com/w3c/automotive/pull/304/commits/86be28eddf9dcb7da791e58ec7fe45de9367060c

As a reminder, this is always easier if commits always deal with only one subject at a time (only changes that are strongly related to each other), and that PRs try to also deal with (mostly) one thing at a time.

PatrickCQ:
Working group decided on 5.11.2019 that this will be reflected in a new PR created with discussion from https://github.com/w3c/automotive/issues/306.

PR #309, opened 2019-10-18
Authentication and Authorization chapter in CORE doc created via input from VIWI. TRANSPORT doc updated to match current VSS modelling.

UlfBj:


spec/Gen2_Core.html
```diff
@@ -553,15 +553,292 @@ <h3>Rules</h2>
...
-        Helpful discussion might be:<br>
-        Authentication: <a href="http://www.w3.org/2018/04/19-auto-minutes.html#item03">http://www.w3.org/2018/04/19-auto-minutes.html#item03</a><br>
-        Volvo assesment: <a href="https://lists.w3.org/Archives/Public/public-automotive/2018Aug/0002.html">https://lists.w3.org/Archives/Public/public-automotive/2018Aug/0002.html</a>, <a href="https://www.w3.org/2018/09/25-auto-minutes.html#item01">https://www.w3.org/2018/09/25-auto-minutes.html#item01</a>
-      </div>
+    <section id="authorization">
+<h2 id="-a-name-userauth-a-user-authentication-and-authorization"><a id="userauth"></a>Authentication and Authorization</h2>
+<p>In order to secure API access and to guarantee delivery of authorized content only, a token-based mechanism (cmp. <a href="https://tools.ietf.org/html/rfc6749">OAuth 2.0 - RFC6749</a>, <a href="https://tools.ietf.org/html/rfc7519">JWT - RFC7519</a>, <a href="https://tools.ietf.org/html/rfc7523">JWT OAuth Profile - RFC7523</a>) is used. OAuth grants an application access to protected data only for specific use cases (scopes) and often for a limited amount of time.</p>
```
danielwilms, line 558: I guess it comes later or was discussed already. Why is OAuth an example only and not used right away?

spec/Gen2_Core.html
```diff
@@ -553,15 +553,292 @@ <h3>Rules</h2>
...
+<td>1</td>
+<td>Physical</td>
+<td></td><td></td>
+</tr>
+</tbody>
+</table>
+<h4 id="firewall">Firewall</h4>
+<p>A firewall shall ONLY let the traffic pass it is configured for. All other ports shall be blocked by default.</p>
+<h4 id="transport-layer-security">Transport Layer Security</h4>
+<p>To secure the transfer of messages and tokens, transport layer security shall be established through the TLS protocol. For more detailed requirements on the required TLS version and cipher suites see Section #ref. Whether a concrete interface needs to be encrypted is subject of the security specification. However, all requests that include tokens MUST be made through an encrypted TLS channel.
```
danielwilms, line 638: > transport layer security shall be established through the TLS protocol

Open question, shall we name TLS as the only solution allowed. Just thinking about QUIC here.

spec/Gen2_Core.html
```diff
@@ -553,15 +553,292 @@ <h3>Rules</h2>
...
+<p>There shall be two types of scopes: <code>internal</code> scopes are granted based on client properties (e.g. client identity or client type) without user interaction; conversely, <code>user</code> scopes may be granted through user interaction, i.e. a popup asking the user to decide whether or not a certain client should be able to use certain functionality. These user-scopes shall be defined in a way that facilitates understanding of a grant decision when being presented to a user, i.e., so that the set of services grouped into the scope carries meaning to a user. Otherwise, users are not able to effectively exercise the control that is expected by the system to protect them from unwanted accesses.</p>
+<p>Every client has to know which scopes to request access á priori, i.e. the scope usually are to be known at implementation time.</p>
+<p><strong>User Scopes</strong></p>
+<p>As a reference, the following set of scopes shall be used to govern access:</p>
+<table>
+<thead>
+<tr>
+<th>Scope Label</th>
+<th>Description</th>
+</tr>
```
danielwilms, line 743: How do the scopes relate to the Taxonomy? Should it be mentioned here or come straight from the taxonomy (e.g. on 1st branch level or similar?

peterMelco:
Hi,
Is there a better way of reviewing changes in the html files than just using GitHub ? I feel it is very tedious … I guess you need to clone the fork and the open the files but with a diff tool that renders html ?

danielwilms (approved):
Issues, which should be listed:
* oauth role
* TLS
* Scoping

Will be addressed after merge

spec/Gen2_Core.html
```diff
@@ -194,6 +192,20 @@ <h3><dfn>stream</dfn></h3>
...
+          </p>
+          <div class="issue">to be described more</div>
+        </section>		
+        <section data-dfn-for="rbranch">
+          <h3><dfn>rbranch</dfn></h3>
+          <p>
+          A node under an <a>rbranch</a> MUST be an <a>element</a>.
+          </p>
+          <div class="issue">to be described more</div>
+        </section>
```
PatrickCQ, line 208: This is not part of the described intend of adding authentication and authorization.

spec/Gen2_Core.html
gunnarx: IMHO a specification _should_ make explicit choices if this is needed to make interoperable solutions.  (If it's not very obvious that it is going to be obsolete within short time)
If a newer technology comes out, it is quite easy to make an updated version of the specification that makes the change.

In fact, it might be useful to consult the typical knowledge out there and go one step further, to specify a minimum version of TLS since there have been security issues found in older versions?)

gunnarx:
I would like to propose, if the group accepts, that rebase (the option named rebase-and-merge in GitHub) is preferred over the merge option.  Otherwise we end up in situations like the above, where the pull request itself has some local merges (how?), and basically 3 of the PR commits are merges themselves.  This PR really only has 3 commits that propose new content, and 3 "useless" merge commits.  It is messy and IMHO the commit content matters, but the PR references you get from a merge commit are only useful to keep while reviewing, and then immediately obsolete.

Issue #306, opened 2019-06-12
Data Model confusion


danielwilms:
Hi,

I promised to summarize the whiteboard discussion, we had at the F2F. One main issue was the misunderstanding around the term data model. I tried to draw a picture on my understanding, excluded the term data model and tried to do some clarification in the terminology, so that we can clearly name, what we're talking about. I think it's crucially important that we get an alignment on this. So @UlfBj, @gunnarx, @tguild, @PatrickCQ, @klotzbenjamin, @rtroncy please have a look and let me know what you think.

![gen2_taxonomy](https://user-images.githubusercontent.com/2179688/59343953-3da7a900-8d0d-11e9-856a-b9ea5fe1efc4.jpg)

gunnarx:
**Main content**

I feel like we are missing something.  I thought as first that **1 - Content**, was the actual concrete database of signal definition, and I felt then that whole description was understandable.  But looking closer "1" is described as a "simple graph" and in the example as a "simple sketch", so it is not the formal _detailed_ database of specific signal names and attributes then?   Where is that covered in the above description?

I don't fully know what the simple tree/sketch is for, and I feel like this needs something like a **4 - Data Definition**, with an arrow  coming from the specification to "4" which is tagged  "instance of" or "implements", or maybe there is a better word.

I was actually about to say that 1, 2, 3 in the examples do not relate to 1, 2, 3 in the previous Domain Taxonomy and should therefore not use the same numbers, but it seems you meant them to be examples? 

At least I think Your "2" in Domain Taxonomy is a specification defining the syntax rules.   But your "2" in Examples looks to me like an actual data definition, not a specification of syntax rules.  The examplelooks like something that may adhere to the rules of a specification, but it *is not* the specification?  Or how do you see it?

**Language and nitpicks**

This is not so much about the concepts, which we must fix first, but more to improve and make it even more clear.

--------

> Domain knowledge expressed in a domain taxonomy to bridge human and machine understanding.

How about "Domain knowledge \***is**\* expressed... 

----
To keep things as short and precise, let's work on conciseness:

* "logical and understandable" can be shortened to either "logical" or "understandable"

----
> The content is the agreement of the taxonomy

I feel this sentence can be clarified, but maybe that follows from my feedback on the concept above.

----
> how to formalize leafs and branches

leafs/leaves
Hmm... https://www.englishforums.com/English/LeafsVsLeaves/nvwxm/post.htm
[...]

PR #311, opened 2019-10-23
Data model chapter updated in CORE doc. Message structure for supported protocols updated in TRANSPORT doc.

UlfBj:


spec/Gen2_Core.html
```diff
@@ -83,18 +83,8 @@ <h2>Terminology</h2>
...
-		This <a href="#tree-example">example</a> shows that major vehicle structures (e.g. body, engine) are near the top of the tree and the logical assemblies and components that comprise them, are defined as their child nodes.
-		Each of the child nodes in the tree is further decomposed into its logical constituents, 
-		and the process is repeated until leafs are reached.
-		A composable node MUST be either a <a>branch</a> or an <a>rbranch</a>.
-		A leaf at the end of a <a>branch</a> cannot be decomposed because and will have an information stored in e.g. a <a>sensor</a>, <a>actuator</a>, <a>stream</a> or <a>attribute</a>.
-		A leaf under an <a>rbranch</a> MUST be an <a>element</a>.
+		The data model used in this specification is the <a href="https://github.com/GENIVI/vehicle_signal_specification">Vehicle Signal Specification</a>, version 2.0 (VSS 2.0). This data model uses a tree-like logical taxonomy to represent the vehicle data. An illustrative example of such a tree structure is shown in <a href="#tree-example">Figure 1</a>. For more details, see the <a href="http://genivi.github.io/vehicle_signal_specification/">VSS 2.0 documentation</a>.
```
PatrickCQ, line 86: if we use an external specification this must be final by the time we release our specification. can that be assured?
link to a specific version needed.

spec/Gen2_Core.html
```diff
@@ -172,203 +162,27 @@ <h2>Data model</h2>
 				
 			@enduml
 			'>
-        <figcaption>Diagram showing an example tree.</figcaption>
+        <figcaption>Diagram showing an example VSS 2.0 tree.</figcaption>
```
PatrickCQ, line 165: are more data models available next to vss 2.0?

spec/Gen2_Core.html
gunnarx: The specification could point to a git tag URL if there is a definite version.  I would say the tag would however not be locked down until we a nearing a Gen2 W3C protocol release, since there are some open improvement proposals.
 
Regarding when the VSS release is considered ready, it is quite a lot into the hands of the same people as are participating in the W3C work (for example the VSS maintainers are participating in the W3C automotive working group).  In other words, the W3C working group could both evaluate the maturity of the current state to see if it is appropriate, and is also essentially responsible for the assurance of it moving to a labeled release.

spec/Gen2_Core.html
UlfBj: In PR #311 it says that Gen2 uses VSS 2.0, so to answer your question, no. 
In VISS it was mentioned that any data model that can express the address to a service using a VSS path expression, which in Gen2 is a URI expression, can be used. I believe that could apply here also.

tguild: As a reminder, Gen2 per charter and past discussions is meant to be usable for other data models besides VSS. We only have one in front of us and while it can be used in spec as an example, need to make clear that is not the only one being supported.

tguild: Actually I stand corrected, @danielwilms took a pass at representing media but within VSS. It would probably be better split out of VSS as a standalone. Gen2 could be used for additional data models defined elsewhere and not under one umbrella. 

https://github.com/GENIVI/vehicle_signal_specification/blob/master/spec/Media/Media.vspec

petervolvowinz: I agree with this - to split out of VSS . It raises potential other issues:  One issue at hand is if the model needs to adhere to "...data model that can express the address to a service using a VSS path expression..." since this reflects the actual API definition making the data model in VIWI incompatible (basically) with the spec. How can we then allow for models other the VSS in this spec that are not a VSS path expressions ? 

I think there are other use-cases apart from media where a dynamic model such as the one presented by VIWI are a much better fit than VSS, but should we really mix these in the same spec ?

UlfBj: I agree that the rbranch and element node types in VSS is a bit confusing, it was introduced to enable porting of media data from VIWI. If that is not to happen, then it is probably better to scrap it. 
The VSS 2.0 path is in my view a URI (segment delimiter instead of VSS 1.0 dot delimiter), so any data model addressing its services through URIs should be applicable.

danielwilms: > How can we then allow for models other the VSS in this spec that are not a VSS path expressions ?

In issue #306 we tried to come up with a theoretical way on how to do it. There was a request to come up with an example, but we never got there. I think media would be an obvious choice, because we talked about it a lot.

I think we should describe a supported taxonomy for Gen2 with a clear documentation on how to extend it (like private branches in VSS), as this is in my opinion a core asset the group can come up with and differentiate to other APIs out there, which try to achieve the same thing.

UlfBj:
I find the rewrite of the Data model chapter introduction to be good.

peterMelco:
So, merging this one now. Issues can of course be raised to suggest further changes

PR #323, opened 2020-03-04
General update of TRANSPORT chapters.

UlfBj:


UlfBj:
For direct browser viewing:
https://rawcdn.githack.com/UlfBj/automotive/36e0faacdc9575c0d0e52bcd17b298aac32e4f35/spec/Gen2_Transport.html

spec/Gen2_Transport.html
```diff
@@ -36,176 +36,195 @@
...
-            <td>There is no client-side fix/solution for this kind of error. If the remote component expected to be reachable after a delay in this power-cycle then server sends a response with status code 503.</td>
-          </tr>
-          </table>
-        </tbody>
+	<p>The server implementation <em class="rfc2119" title="SHALL">SHALL</em> support at least the error numbers and reasons listed in the table below.</p>
+        <p>The client <em class="rfc2119" title="SHOULD">SHOULD</em> support any status code defined in <a href='#bib-HTTP-PROTOCOL'>[RFC2616]</a>.</p>
+<table id="errorDefs" class="parameters">
+	  <tbody><tr>
+	    <th>Error&nbsp;Number&nbsp;(Code)</th>
+	    <th>Error Reason</th>
```
gunnarx, line 74: Is "Error Reason" a reference to "Reason-Phrase" in RFC 2616?

Is it expected that the exact text given here, for example "filter_invalid" will be provided in the reason phrase?  Or is the error reason communicated some other way?

I think the specification needs to be more formal by referring to the exact same concept name as the RFC, and/or specify that these texts will be written in the Reason-Phrase.   In particular since, as far as I understand, these will need to be machine processed by the client so that it can write out the "Error Message", but RFC 2616 says "the Reason-Phrase is intended for the human user" and the code only is for machine processing?

spec/Gen2_Transport.html
```diff
@@ -224,15 +243,26 @@ <h2>Read</h2>
...
+              }<br><br>
+              Error response:<br>
+              HTTP/1.1 404 Not Found<br>
+              Content-Type: application/json; charset=utf-8<br>
+              Date: Tue, 13 Jun 2014 19:47:27 GMT<br>
+              Connection: keep-alive<br>
+              Transfer-Encoding: chunked<br>
+              {<br>
+              <blockquote>
+                {"message": "No match for path in tree."}<br>
```
gunnarx, line 261: Is this body content required, or optional.  Is it in addition to the error code and or error reason?

spec/Gen2_Transport.html
```diff
@@ -271,10 +301,43 @@ <h2>Service discovery read</h2>
               }<br>
             </p>
           </section>
+          <section id="https-search-read">
+            <h2>Search read</h2>
+            <p>
+              <b>Example:</b><br>
+              Request:<br>
+              GET Vehicle/Cabin/Door?$path=*.*.IsOpen HTTP/1.1<br>
```
gunnarx, line 309: Think I might have mentioned before that I think wildcards in the URL is more straight forward, but I don't know what the web philosophers think about it?

**Proposal 1 (in PR):**
```
GET Vehicle/Cabin/Door?$path=*.*.IsOpen 
```

**Proposal 2, with wildcards in URL**
```
GET Vehicle/Cabin/Door/*/*/IsOpen 
```

But if the \*/\* in URL is not considered OK (because it does not address a specific "document" in terms of web philosophy) and a query is desirable, then perhaps instead of splitting the path it could make more sense to ask for a generic resource like SignalQuery and put the entire path in the query:

**Proposal 3:**
```
GET /SignalQuery?$path=Vehicle.Cabin.Door.*.*.IsOpen
```

(My order of preference: Proposal 2, Proposal 3, Proposal 1)

spec/Gen2_Transport.html
```diff
@@ -271,10 +301,43 @@ <h2>Service discovery read</h2>
...
+              </blockquote>
+              }<br>
+            </p>
+          </section>
           </section>
-          <section data-dfn-for="Foo">
+          <section data-dfn-for="https-update">
             <h2>Update</h2>
             <p>
+The client may request that the server sets the value of one or more signals e.g. to lock one or more doors or open a window by sending an HTTPS POST request to the server. If the server is able to satisfy the request its response SHALL have a 200 OK status code set. If an error occurs e.g. because the client is not authorized to set the requested value, or the value is read-only, the server response SHALL have the status code set to indicate error.
```
gunnarx, line 338: > "e.g. to lock one or more doors or open a window by sending an HTTPS POST request to the server."


The purpose of setting a signal could be to actuate a function but I don't think it always is.  

Is it also possible (?) that the web protocol is used by the source of a signal to just set its current measured value.  The part of the system that has measured the status or value of something could indicate this to subscribers by writing (publishing) the signal through the web protocol?

I am as you might know _accepting_ the idea of some actuators as writable signals, but generally more inclined to prefer an explicit API for command-and-control situations.  I wonder therefore which should be highlighted in the specification, or even if it should be highlighted.  One  alternative is to not indicate anything in this general spec about the expected behavior of writing to a signal.  
(Case in point - which signal would the window-opening signal be in VSS today?   Is it one of the "*.IsLocked" or "IsOpen" signal?  I'm a bit uncomfortable with assuming that writing a status signal like that will actually cause the window or door to move, although I suppose this is clarified by the type: actuator...

spec/Gen2_Transport.html
```diff
@@ -318,50 +381,183 @@ <h2>Authorized update</h2>
...
+	instance may either be instantiated natively or be created using a
+	'standards compliant' WebSocket JavaScript library.</p>
+	<p>A WebSocket can also be initiated from a native (e.g. C++) Application
+	or from an Application written using a 'Managed Runtime' language like
+	Java or C#. It is assumed that native and managed clients use a
+	suitable standards compliant WebSocket library to request that a
+	WebSocket connection is opened on the server.</p>
+
+	<p>Implementations that support additional devices or multiple Gen2 services
+        should provide discovery. Alternatively, the location of a particular VIS
```
gunnarx: What is VIS?   Do we use this name officially?  Or misspelling of VISS?  (and do we still use that name or not?)  It's not defined I think, or?

spec/Gen2_Transport.html
```diff
@@ -318,50 +381,183 @@ <h2>Authorized update</h2>
...
+        Server instance on the local vehicle network may be handled by
+        configuration, either as part of a package manifest or by consulting a
+        registry on application install. The 'wwwgen2' hostname in this
+        specification is used an example.
+	</p>
+	      
+	<p>A client running on the vehicle is able to connect to the
+	Gen2 Server instance using the hostname e.g. 'wwwgen2' and uses the
```
gunnarx, line 411: Same question... Is "Gen2" an official name meaning something?  I know we use it, but it's time to become more formal about this I think?

spec/Gen2_Transport.html
```diff
@@ -318,50 +381,183 @@ <h2>Authorized update</h2>
...
+	default port 443. The hostname 'wwwgen2' may locally be mapped to the localhost
+	IP address 127.0.0.1 e.g. by adding an entry to the /etc/hosts file.</p>
+
+	<p>The sub-protocol name <em class="rfc2119" title="SHALL">SHALL</em> be 'wvss' with a version number suffix, e.g. wvss2.0. The sub-protocol version will
+	be associated with exactly one Vehicle Server Specification (VSS) version so that the client and server can 
+	correctly validate and parse request and response message packets.</p>
```
gunnarx, line 417: I'd like to clarify what "exactly one VSS version" means here.

Is it OK to have minor or patch release upgrades, for example?  Those would presumably be compatible in format, but changed their content (signal list) slightly?

And I imagine the defined protocol version, which is very important to agree on in this exchange, might also be used with some proprietary extensions of the VSS signal list?

So, is the intention to lock down the _format_ of the VSS, or also content?  Example of incompatible format changes could be extensions that have been made in version 2,  such as instances [1..2], arrays, etc.  These VSS _capabilities_ must be negotiated by the protocol first of all.  But is there also more?

spec/Gen2_Transport.html
```diff
@@ -318,50 +381,183 @@ <h2>Authorized update</h2>
...
+	<pre class="highlight hljs javascript" aria-busy="false">	  <span class="hljs-keyword">var</span> vehicle  = <span class="hljs-keyword">new</span> WebSocket(<span class="hljs-string">"wss://wwwgen2"</span>, <span class="hljs-string">"wvss2.0"</span>);
+	</pre>
+	<p>The client <em class="rfc2119" title="SHALL">SHALL</em> connect to the server over HTTPS and request that
+	the server opens a WebSocket. All WebSocket communications between
+	the client and server <em class="rfc2119" title="MUST">MUST</em> be over ‘wss’. Non encrypted communication
+	is not supported, hence the server <em class="rfc2119" title="MUST">MUST</em> refuse ‘ws’ connection requests.</p>
+	      
+        <p>This specification assumes that a single WebSocket is used to enable communication between a client application
+        and the server. It is not explicitly prohibited for the client to request that the server opens more than one WebSocket,
```
gunnarx: remove 'explicitly'.
... actually to be more in line with the rest of the specification, why not simply:
"The client MAY open more than one websocket"

I think you can simplify:  Write 
'request more than one socket'
or 'client to open more than one socket'
 instead of 
'request that the server opens more than one socket'

spec/Gen2_Transport.html
```diff
@@ -318,50 +381,183 @@ <h2>Authorized update</h2>
...
+        however, the server <em class="rfc2119" title="MAY">MAY</em> refuse to open a subsequent WebSocket connection and the client is responsible for handling this
```
gunnarx: 'However' starts a new sentence.

spec/Gen2_Transport.html
```diff
@@ -318,50 +381,183 @@ <h2>Authorized update</h2>
...
+        that the server implements a simple concurrency model where lost
+        updates and dirty reads could potentially occur if the server has
+        more than one WebSocket connection open.</p>	 
           </section>
-          <section data-dfn-for="Foo">
-            <h2>Closure</h2>
+          <section data-dfn-for="wss-closure">
+            <!--OddPage--><h2>Closure</h2>
+	<p>The WebSocket may be closed by either the client or the
+	server by invoking the ‘close()’ method on the WebSocket
```
gunnarx, line 448: This is slightly weird, since it specifies what can be done using javascript instead of on the HTTP protocol level.  But maybe that's OK and we are doing that in other places too?.   Do we have a principle for this in the whole specification.    Javascript is only one possible implementation language.  On the other hand we could write at the top that we will use Javascript for some of the examples, and stick to that...

spec/Gen2_Transport.html
```diff
@@ -318,50 +381,183 @@ <h2>Authorized update</h2>
...
+</p>
+<br>
+<table class="simple">
+<thead>
+  <tr><th>Object Name</th><th>Attribute</th><th>Type</th><th>Required</th></tr>
+</thead>
+<tbody>
+  <tr>
+    <th rowspan="5"><dfn data-dfn-type="dfn" id="dfn-getrequest">getRequest</dfn></th>
+    </tr><tr><td><a href="#dfn-action">action</a></td><td><a href="#action-def">Action</a></td><td>Yes</td></tr>
```
gunnarx, line 483: (continuing review... I had to start a new one because it was already submitted)
Instead of this having a link to "Action" (which means any action) why not specify that the action must equal "get"?   For a getRequest, the value MUST be "get", right?

Same below for the two other message types.

spec/Gen2_Transport.html
```diff
@@ -318,50 +381,183 @@ <h2>Authorized update</h2>
...
+  <tr>
+    <th rowspan="5"><dfn data-dfn-type="dfn" id="dfn-getsuccessresponse">getSuccessResponse</dfn></th>
+    </tr><tr><td><a href="#dfn-action">action</a></td><td><a href="#action-def">Action</a></td><td>Yes</td></tr>
+    <tr><td><a href="#dfn-requestid">requestId</a></td><td>string</td><td>Yes</td></tr>
+    <tr><td><a href="#dfn-value">value</a></td><td>object</td><td>Yes</td></tr>
+    <tr><td><a href="#dfn-timestamp">timestamp</a></td><td>integer</td><td>Yes</td></tr>
+  
+</tbody>
+</table>
+<p>In the table above the value attribute has the type object as it may contain multiple values, each associated to other attributes. However, an atomic value representation MUST be of type string. </p>
```
gunnarx: Not 100% sure I understand this text.  Is there a clearer way to say it?

spec/Gen2_Transport.html
```diff
@@ -318,50 +381,183 @@ <h2>Authorized update</h2>
...
-                "value": 2372,<br>
+                "error": {"number": 404, "reason": "Not found", "message": "No match for path in tree."},<br>
                 "timestamp": 1489985044000<br>
               </blockquote>
               }<br>
             </p>
-          <section data-dfn-for="read-token">
+          <section data-dfn-for="wss-authorized-read">
             <h2>Authorized read</h2>
+            <p>In the case a VSS node that is addressed by the path contains access control metadata,
```
gunnarx: Since we have not decided how to represent this I would propose:
"In the case the operation on the VSS node that is addressed requires authorization..."

and then "In the case" can be simplified to "If"

spec/Gen2_Transport.html
```diff
@@ -370,68 +566,158 @@ <h2>Authorized read</h2>
...
               </blockquote>
               }<br>
             </p>
           </section>
-          <section id="ws-service-discovery-read">
+          <section id="wss-service-discovery-read">
             <h2>Service discovery read</h2>
+            <p>A client may issue a service discovery read request to access the metadata defining the nodes in the VSS tree. 
+               A successful response will contain the metadata from all nodes of the subtree defined by the subtree root node that is addressed by the path.
+               The service discovery read request MUST contain a query component with the reserved word "$spec", and an associated integer value (c.f. CORE spec, "Filtering" chapter).
```
gunnarx: c.f. --> [cf.](https://en.wikipedia.org/wiki/Cf.) ?

gunnarx: I'm not too familiar with the usage of cf. so I looked it up.  It seems the Wikipedia article suggests it should only be used if you are comparing to something different.  Here it is a reference to more details in support of the statement.

spec/Gen2_Transport.html
```diff
@@ -370,68 +566,158 @@ <h2>Authorized read</h2>
...
+                "timestamp": 1489985044000<br>
+              </blockquote>
+              }<br>
+            </p>
+          </section>
           </section>
-          <section data-dfn-for="Foo">
+          <section data-dfn-for="wss-update">
             <h2>Update</h2>
+            <p>The client may request that the server sets the value of one or more signals e.g. to lock one or more doors or open a window by sending a <a href="#dfn-setrequest" class="internalDFN" data-link-type="dfn">setRequest</a> message to the server. If the server is able to satisfy the request it <em class="rfc2119" title="SHALL">SHALL</em> return a
```
gunnarx, line 634: _See comment on previous equivalent chapter_

spec/Gen2_Transport.html
```diff
@@ -370,68 +566,158 @@ <h2>Authorized read</h2>
...
                 "timestamp": 1489985044000<br>
               </blockquote>
               }<br>
             </p>
           </section>
+          <section id="wss-search-read">
+            <h2>Search read</h2>
+            <p>A client may issue a search read request to access multiple values in one request message. 
+               A successful response will contain the values from all nodes matching the path concatenated with $path value (a delimiter is automatically inserted). 
+               The search read request MUST contain a query component with the reserved word "$path", and an associated extension to the path (c.f. CORE spec, "Filtering" chapter).
```
gunnarx: _same_

spec/Gen2_Transport.html
```diff
@@ -370,68 +566,158 @@ <h2>Authorized read</h2>
...
+</table>
+<br>
+<table class="simple">
+<thead>
+  <tr><th>Object Name</th><th>Attribute</th><th>Type</th><th>Required</th></tr>
+</thead>
+<tbody>
+  <tr>
+    <th rowspan="5"><dfn data-dfn-type="dfn" id="dfn-seterrorresponse">setErrorResponse</dfn></th>
+    </tr><tr><td><a href="#dfn-action">action</a></td><td><a href="#action-def">Action</a></td><td>Yes</td></tr>
```
gunnarx, line 674: See comment on previous equivalent chapter

spec/Gen2_Transport.html
```diff
@@ -370,68 +566,158 @@ <h2>Authorized read</h2>
...
                 "requestId": "5687",<br>
+                "error": {"number": 404, "reason": "Not found", "message": "No match for path in tree."},<br>
                 "timestamp": 1489985044000<br>
               </blockquote>
               }<br>
-            </p>
-          <section data-dfn-for="Foo">
+          <section data-dfn-for="wss-authorized-update">
             <h2>Authorized update</h2>
+            <p>In the case a VSS node that is addressed by the path contains access control metadata,
```
gunnarx: _same_

spec/Gen2_Transport.html
```diff
@@ -451,54 +737,129 @@ <h2>Authorized update</h2>
...
                 "requestId": "6578",<br>
                 "timestamp": 1489985044000<br>
               </blockquote
               }<br><br>
+              Error response:<br>
+              {<br>
+              <blockquote>
+                "action": "subscribe",<br>
+                "requestId": "6578",<br>
+                "error": {"number": 404, "reason": "Not found", "message": "No match for path in tree."},<br>
```
gunnarx, line 827: Is there not redundancy in providing number, reason and message?
In this particular case, number and reason are equivalent (I think?), meaning that 404 **means** "not found", right?

gunnarx:
I did not repeat all of the comments, please check if something is changed that it is changed in every place that is similar...

spec/Gen2_Transport.html
UlfBj: I believe so. The Error type is inherited from VISS. 
A reference to RFC2616 for the error number and error reason is added to the error definitions chapter.

spec/Gen2_Transport.html
UlfBj: It is in the error definitions chapter declared as mandatory.

spec/Gen2_Transport.html
UlfBj: I put my vote behind proposal 1.

spec/Gen2_Transport.html
UlfBj: I put my penny on that it is
Vehicle.Cabin.Door.Row1.Left.Window.Position
The type, sensor or actuator, should help clarify.

spec/Gen2_Transport.html
UlfBj: cut and paste error from VISS spec. VIS changed to Gen2.

spec/Gen2_Transport.html
UlfBj: Gen2 is the working name for the spec, decided at some WG meeting in the early stages of this charter.

spec/Gen2_Transport.html
UlfBj: VSS version is another typo. Changed to Gen2 version. 
I think this version issue you point at is important, and needs more thinking. 
I propose that an Issue is created on this, so that it can be worked on without putting this entire PR on hold. 
Issue #325 created.

spec/Gen2_Transport.html
UlfBj: Changed to: The client MAY open more than one websocket

spec/Gen2_Transport.html
UlfBj: Updated.

spec/Gen2_Transport.html
UlfBj: Inherited from VISS through the noble art of copy and paste. 
If there is consensus we shall not follow this legacy principle, we can change.

spec/Gen2_Transport.html
UlfBj: VISS legacy. 
The table column is Type, and Action is the type here.

spec/Gen2_Transport.html
UlfBj: An attempt to make it clearer is created.

spec/Gen2_Transport.html
UlfBj: Done. On all three occasions of it.

spec/Gen2_Transport.html
UlfBj: Thanks for educating me:).
Now replaced by "see" on all occasions.

spec/Gen2_Transport.html
UlfBj: VISS legacy. 
Issue #326 is created for WG to discuss, in order to not hang the PR.

UlfBj:
PR is updated as per comments from @gunnarx .

tguild:
merging per WG call and convention

PR #324, opened 2020-03-24
version

tguild:
Draft phrasing to explain VSS versioning

### DIFF
--- a/vehicle_data/vehicle_information_service.html
+++ b/vehicle_data/vehicle_information_service.html
@@ -175,6 +175,11 @@ Working Draft of this specification</a>, the following <a href="viss-diff.html">
 
       <p>This specification defines a number of methods for accessing vehicle data which are strictly agnostic to the data model. Any data model where data and signals can be specified using a string could potentially be supported. However this version of the specification specifies that the data model is the GENIVI Vehicle Signal Specification [[!VSS]]. The VSS supports both extensibility and the ability to define private branches and will be used for each of the examples within this specification.</p>
 
+      <p>The version of VSS is made clear within VSS
+      itself as of version 2.0. In the absence of VersionVSS.Major,
+      VersionVSS.Minor, VersionVSS.Patch and VersionVSS.Label it
+      should be assumed VSS is version 1.0.</p>
+      
       <p>In addition, the 'tree' of signals that is accessible at any point in time may also vary depending on standard access control principles.
          That is, it can vary based on the identity of the user (person or organisation) requesting the data and/or the application or 
 	 device (e.g. vehicle) where the request originates.</p>


danielwilms (approved):

UlfBj:
I think this PR should include, or be accompanied by another PR, the addition of the VersionVSS nodes to the YAML spec. If not explicitly added to the tree (in case this is supposed to be added in the "deployment phase"), the information about where in the tree this should be added, and the node declarations should be defined.

tguild:
@gunnarx 
Would it be possible to take the the zipped archive @danielwilms made, extract (portions?) and publish with a permanent genivi.org URI? I would rather point to VSS spec directly instead of a zipfile. vss_rel_1.0.json and README.md (I can create rendered html) at the least, YAML and rest of the contents can remain in zip.

https://github.com/GENIVI/vehicle_signal_specification/releases/tag/v1.0

gunnarx:
* When a release is created, then there is an automatic archive created (zip, tar.gz) containing the source code from the repository, but it is also possible to upload **additional files** that are stored along side the archives.  @danielwilms could look into that.  I don't know if it is possible to edit an old release like 1.0 or not.  I could see it was possible to upload while creating a new release at least, but Daniel would have to check the old one.

* But to do what you want, it is also possible to point to any file in the GitHub repository using a stable link, by specifying a commit hash or branch, or tag.   Here is a link to a file shown in the version that belongs to the v1.0 tag:

https://github.com/GENIVI/vehicle_signal_specification/blob/v1.0/spec/VehicleSignalSpecification.vspec
or you can get the raw contents too:
https://raw.githubusercontent.com/GENIVI/vehicle_signal_specification/v1.0/spec/VehicleSignalSpecification.vspec

If one of these options are acceptable so that we can use GitHub only, I would find this to be the best option.

However, before we do.   I'm not yet certain that pointing to the single file VehicleSignalSpecification.vspec is quite the right process.  I think we should discuss this in the VSS project because that file is generated from all the smaller files in the directory structure:
https://github.com/GENIVI/vehicle_signal_specification/tree/v1.0/spec
and...a good principle is to only have one source of truth, so there is no risk of misalignment.

For the moment, I believe it is the individual smaller files in combination that make up the official VSS, but that might just be my opinion.  I think this is worth discussing.

gunnarx:
@UlfBj Your comment above, is that covered already by this PR
https://github.com/GENIVI/vehicle_signal_specification/pull/136?
If not, I'm not sure I understood the comment.

gunnarx (approved):

UlfBj:
>> @UlfBj Your comment above, is that covered already by this PR
GENIVI/vehicle_signal_specification#136?

This had escaped my memory. Forget my comment.

PR #329, opened 2020-06-02
Timestamp format changed to ISO8601.

UlfBj:


tguild (approved):

tguild (approved):
'with subsecond resolution as an acceptable alternative' wording seems wrong. subsecond is not an alternative to seconds but optional precision. Perhaps instead:

'with subsecond resolution as an optional degree of precision when desired.'

Unix timestamp for tokens what oauth uses?

tguild (approved):

UlfBj:
I will update the subsecond sentence according to your proposal, it will come in a following PR, which is already stacked up on my fork. I do not know how to bypass that with this update.

>>Unix timestamp for tokens what oauth uses?

It is at least recommended to minimize token size.

PR #334, opened 2020-08-04
Access control chapter rewrite

UlfBj:


peterMelco:
So, one general comment/question is how we use capital letters within sentences ? 

Ex:
"In the Short term flow the Client must contact the AGS every time it needs a new Grant token"

I guess the purpose is to make emphasize on certain words ?  I don't see why we need to define our own way of capitalization though.

UlfBj:
The idea was to distinguish a Gen2 defined Client from a common client. But
capitalizing on that concept may lead to confusion.
i am fine with either way.
br
ulf

spec/Gen2_Core.html
```diff
@@ -343,78 +275,250 @@ <h3>Filter combinations</h2>
...
+          <p>
+          The request shall contain the first two parameters below, the other is optional:
+          <ul>
+            <li>AG token: A signed token with claims needed for the validation of the Client request.</li>
+            <li>Purpose: The Purpose shall be one of the short text entries from the Purpose list.</li>
+            <li>Proof: A proof of possession of the public key in the AG token.</li>
+          </ul> 
+            All access token request shall include an AG token. Moreover, if the Access Grant token includes a public key, 
+            the Client should also include in the request a fresh proof of possession for this public key. 
+            This shall be a fresh timestamp that is digitally signed using the associated private key.
```
SebastianSchildt: I do not quite get the concept of "proof of possession", in this document  as there is no such thing as a "fresh" timestamp (it is just a number, and one that can be easily predicted). And why it mentions "public" key, as I can not really "posses" it (hence public). To prove I have a corresponding private key, the standard way would be to sign a nonce (so basically a _random_ number not decided by me) received by the entity that wants me to prove something.

Maybe I just did not get the point completely, but in that case I still hope it can be rewritten to make it clear to people like me 😄

UlfBj: Possession relates to the key pair, maybe that should be more clear. 
A challenge response protocol where the ATS presents a nonce to be signed by the Client would be somewhat more robust, but adds complexity to the interactions. We think that the signing of a "timestamp" that must not be more than maybe a few mins off from system time should provide sufficient security. It can of course be signed at any earlier time, but why would the Client want to do that?

SebastianSchildt: Understood, so the intent is really proving access to the secret (key) corresponding to the public key. 

I am no security guy, but it feels wrong to me, to "proof" something, without a challenge. At least this would trivially open a window for replay attacks, as you can not expect the clocks to be synchronised perfectly, so the receiver must allow for some range of accepted timestamps. Wouldn't just using the private key (I mean for whatever it is used for, not just for signing the timestamp) already be proof, and what would be the attack vector here, without the "proof"? I.e. if I am the "wrong" client, or I lost access to the private key. What would proving possesion help in this step?

UlfBj: Yes, the accepted clock unsynchronization can be used for replay attacks if an attacker can get hold of a copy of both token and proof. This communication is encapsulated by HTTPS usage, and the proof is not a persistent data item (like the token). 
So it is the simplified protocol vs the risk of a replay attack. Can be discussed, absolutely. 
Without a proof, an attacker "only" needs to obtain the token, a considerably easier task.

SebastianSchildt: But in that case, when considering "stealing the token" as attack vector, it implies I can not send a token unencrypted anyway. So wouldn't the/a public key in the token refer to the my TLS client key, thus I have already proved who I am by establishing the connection? Even if we are not demanding TLS, basically we want an "authentication" step here, correct?  And as there are countless protocols and mechanisms to do authentication, it seems dangerous to roll your own? Or is the timestamp signing some common/accepted measure? If so, can you reference it?

tguild:
It might make more sense to have the 6.4 Actors section moved up to 6.2, following architecture. Somewhat repetitive parts in actor/protocol parts and details belong more in protocol part, suggest shortening actors descriptions.

Client Context, Proof and Purpose should perhaps be explained early on before referenced.

6.2 is it realistic to expect an application to be able to generate its own keys? If provided by the host environment, invoking commands and key storage may need an API itself. Consider instead a number of pre-issued pairs?

6.3.1 - Is VIN necessary? I can see apps permitted to run on a vehicle that would not need to know this potentially personal identifying information. VIN itself can be used to determine year, make, model, country built, potentially trim package, lookup outstanding recalls, crash history (carfax requests) etc. VIN may be a protected signal and not something known or given to all client applications.

This part is unclear or confusing, perhaps from reusing term proof. "If a public key is included in the request, the proof parameter shall also include of proof of possession for this key." I need further understanding of proof mechanisms.

6.3.2 "depending on the input in the request message" which supplied parameters indicate desired duration for tokem, part of context?

6.4.2 typo s/AG token create a prof of possession/AG token create a proof of possession/

6.4.3 "The expiry time must be later than the current system time of the AT server run-time. To allow for some time synchronization inaccuracy an error of tens of seconds may be allowed." These two statements are in contradiction 'must' be later but may be close enough within margin or error. A better second statement might be "To allow for some time synchronization inaccuracy and minor network latency, expiry times should be set to include potential modest margin of error, possibly as long as tens of seconds."

6.4.4 401 Unauthorized more appropriate than 404 for missing token, what response from http server gives repeatedly if given incorrect or missing credentials. After client gives up on a 401 'loop', 403 instead of 406 may be more appropriate.

6.4.5 should include mechanism for capturing consent is outside scope.

6.4.5 and 6.4.6 belong under actors?

Trying to understand Aud parameter's purpose.

spec/Gen2_Core.html
isaacagudo: Proof-of-Possesion (PoP) using JWT is described in the following RFC: [https://tools.ietf.org/html/rfc7800] (RFC 7800)
The idea is to proof possession of a private key to anyone who learns about the corresponding public key, we might need to review the writing to make that more clear. It is in essence a public key based authentication protocol but you are not probing the ownership of any identity, just a private key.

As Sebastian pointed out, if we use timestamp for the PoP, i.e. a predictable number, the proof is not actually "fresh", it will just show the willingness of the holder of the private key to use it at his particular timestamp, that could be in the future or in the past. We need to change the term fresh if we finally go for the timestamp and consider/explain carefully the attack model here. I think the legit holder is not gaining any advantage producing future PoP and storing them, if the holder does that, it would be a similar approach to S/KEY where you have many one-time-passwords (OTP) to authenticate. The advantage of having different "passwords" would be that an attacker that intercepts one of them could not derive any other one, making the attack very limited in time, i.e. if a new PoP is issued every minute the attacker would only be able to impersonate the holder for a minute, but if the PoP is issued for every request the attack would only be able to authenticate the holder if it manages to reach the Access Token server before the legit holder although the legit holder would detect that becase the server would refuse the legit request.

Other options for the PoP nonce would be:

- A client generated random number. The problem with this approach is that the server has no means to probe that the nonce is legit ...
- A server generate random number. That is the standard pattern for challenge response mechanisms, it works great but it would requiere extra messages ...
- A publicly available random number. There are different initiatives to produce public and veritable randomness. This public number would ensure certain degree of freshness and avoid extra protocol messages. E.g. [https://www.cloudflare.com/leagueofentropy/](Leage of Entropy) and [https://csrc.nist.gov/projects/interoperable-randomness-beacons] (Interoperable Randomness Beacons)

Finally we might need to address the case in which the legit holder is compromised and its private key is stolen. In this case we need to implement some revocation mechanism. For short term tokens revocation is usually not a concern becase the attacker only have a small window for impersonating the client, but requires the client to be online an refresh its tokens. I would only recommend using private keys if they can be hardware protected, where the risk of exposure is very limited, otherwise I would recommend shot term tokens.

The design goal of long term tokens was to provide a solution where the vehicle could be offline for large periods of time without an impact on clients, as a side effect this system is more secure against eavesdropping attacks. We might need to revise whether this is still a priority.

isaacagudo:
> 6.2 is it realistic to expect an application to be able to generate its own keys? If provided by the host environment, invoking commands and key storage may need an API itself. Consider instead a number of pre-issued pairs?

I don't think every application should use the long term tokens, but in any case the OS/host should provide a mechanisms to provision key pairs, I think this is outside of the scope.

> 6.3.1 - Is VIN necessary? I can see apps permitted to run on a vehicle that would not need to know this potentially personal identifying information. VIN itself can be used to determine year, make, model, country built, potentially trim package, lookup outstanding recalls, crash history (carfax requests) etc. VIN may be a protected signal and not something known or given to all client applications.

I think VIN could be a protected resource from a privacy by design perspective, if the application only needs speed why do we need to allow access to the VIN too. This is something we might need to revise.

> This part is unclear or confusing, perhaps from reusing term proof. "If a public key is included in the request, the proof parameter shall also include of proof of possession for this key." I need further understanding of proof mechanisms.

Maybe if you read the previous comment you'll get a more clear picture of PoP

tguild:
merging as agreed on WG call

https://www.w3.org/2020/09/08-auto-minutes.html

PR #341, opened 2020-10-06
Access control chapter updates.

UlfBj:


tguild (approved):
Looks mostly like rearranging sections. I still feel VIN may be something withheld as personal identifying information and should be optional

PR #342, opened 2020-10-08
Version links

### LINKED ISSUE #307: VSS version in VISS
We should update VISS to convey what version of VSS is being exposed. My initial thinking is it would be appropriate to handle this in the getMetaData() method.

tguild:
I added wording earlier regarding version expectations and going over old threads was awaiting, long settled by @danielwilms in GENIVI repo, links to clarify what we can point to as VSS1.0

@peterMelco after approving the pull request, please initiate a Call for Consensus. Here is an old example from Rudi

https://lists.w3.org/Archives/Public/public-automotive/2018Jan/0008.html


tguild:
HTML rendered view for easier reading

https://raw.githack.com/w3c/automotive/gh-pages/vehicle_data/vehicle_information_service.html

PR #350, opened 2020-11-19
Rewrite of Filtering chapter.

### LINKED ISSUE #345: Fragment instead of query in signal discovery requests
* Current syntax for a signal discovery request is
** GET Vehicle/Drivetrain/FuelSystem?$spec=0
* Proposed syntax
** GET Vehicle/Drivetrain/FuelSystem#$spec
*** To get the full tree below
** GET Vehicle/Drivetrain/FuelSystem?$searchEQ*.*#$spec
*** To get two levels
* RFC 3986
** “The query component contains non-hierarchical data that, along with data in the path component, serves to identify a resource within the scope of the URI's scheme and naming authority”
** “The fragment identifier component of a URI allows indirect identification of a secondary resource by reference to a primary resource and additional identifying information.”

UlfBj:


tguild:
https://rawcdn.githack.com/UlfBj/automotive/07ad90dd6d4ed420c35b176eadc63eb4497f7969/spec/Gen2_Core.html

gunnarx:
**Chapter 6:  Metadata Request**

**spec vs status:**

This separates metadata in two categories, VSS-tree defined, and otherwise defined in the system.

* For spec, you can't specify more exactly what you want so you get everything, correct?
* But when it comes to status, you can request something specific (capture-filter)
* Note, for status the comment `****Any more?****` remains.

I feel "status" is not the best indicative word.  

Putting it all together I wonder if **spec** and **status** might be unified instead, and that we allow the request for any individual thing, also for the VSS-tree stuff?

So instead of `/vsspath#metadata-type`  you could write:

```
/vsspath#sample-rate
/vsspath#capture-filter
/vsspath#unit
/vsspath#datatype
/vsspath#description
/vsspath#deprecation
/vsspath#all
```
so basically individually ask for any metadata, regardless if it is `static` (defined in VSS tree) or something unique/dynamic to the system (sample-rate).  or `all` to get everything?
and/or `#static-metadata` for metadata from vss and `dynamic-metadata` for anything extra...

alternatives:
`#metadata` instead of `#all`
...

gunnarx:
**Chapter 5:  Filter Request**.  The technical requirements (behavior) of this chapter seem OK for me.  
If I add feedback it would only be detailed proof-reading to fix a few descriptions to make it slightly clearer.

PR #384, opened 2021-05-05
Fixes some spelling errors

erikbosch:
Noticed some spelling errors. And I guess the first S in VSS shall be "Signal", not "Signals", and that the "This is required" paragraph shall be removed.


erikbosch:
@tguild, I assume this can be considered as non-substantive. Or if you prefer, you can fix them in a separate PR and I can close this one

PR #401, opened 2021-08-25
Issue399

UlfBj:


tguild:
@UlfBj Per conversation on last call

Following your sentence "Please note that for actuators this does not return a target value that might have been set, but not yet fully actuated." consider adding:

"There may be reasons why a target value will not be fully actuated such as exceeding allowed range or race condition with another client."

UlfBj:
@tguild : A newer update of this issue is found on PR#402, please review that.

UlfBj:
Replaced by PR#406

Issue #399, opened 2021-08-04
Issue on Authorized Read in VISSv2-Transport


wonsuk73:
There are read, update, subscribe and unsubscribe operations in Current VISSv2-Transport spec. In addition these operations are seperated with authorized and unauthorized as below.

HTTPS:
5.1.2.1 Read
  5.1.2.1.1 Authorized Read
  5.1.2.1.2 Search Read
  5.1.2.1.3 History Read
  5.1.2.1.4 Service Discovery Read
5.1.2.2 Update
  5.1.2.2.1 Authorized Update

Websocket:
5.2.2.1 Read
  5.2.2.1.1 Authorized Read
  5.2.2.1.2 Search Read
  5.2.2.1.3 History Read
  5.2.2.1.4 Service Discovery Read
5.2.2.2 Update
  5.2.2.2.1 Authorized Update
5.2.2.3 Subscribe
  5.2.2.3.1 Authorized Subscribe
  5.2.2.3.2 Curve Logging Subscribe
5.2.2.4 Unsubscribe

Could we allow a client to access vehicle data without authentication? I guess the car manufacturing company might not want to provide their car data without their control. There might be a few level of authentication to control the data access according to security or privacy level.  Am I right?
If it is right, I would suggest to add the phase for explanation of this context, and to change all of unauthenticated operations with authenticated operations. What's your opinions?

wonsuk73:
@UlfBj, @peterMelco and @tguild Could you review and share your view?

SebastianSchildt:
My (not necessarily correct) understanding at least for WS  was, that "Authorized Read" is the special case, where you can put the whole token into each read request, which might be a good pattern if you want a single value ever hour/day or so, while otherwise it has tremendous overhead as a JWT token likely is  several times the size of the query and the response combined.

I assumed normal reads are only possible after a session using a token has been established. (of course depending on the implementation maybe it opts to allow certain reads without any prior authentication).

UlfBj:
>> Could we allow a client to access vehicle data without authentication? 

We do allow that as the access control is optional. On top of that, even if an OEM supports access control, ch. 8.8 "Access control selection" makes it possible to apply access control to only selected nodes of the tree. So some nodes may then not require a token for access. 

If access control is enabled on a signal, then according to the spec the token MUST be provided in each request to it. 
I do not think we should change that and allow some requests to this signal to be without the token.

SebastianSchildt:
[...]

PR #419, opened 2021-09-15
Sending the JTI instead of the whole access token to save bandwidth

isaacagudo:
As discussed in issue  #399, we want to have the possibility to send a short token to authorize requests to the server. As agreed in the last meeting, sending just the JTI seems like a reasonable solution. 
I tried to change as little as possible in the specification. I could elaborate a bit more on the caching strategy for the server but I think this is out of the scope.


UlfBj (commented):
A requirement on the server that the token must be cached and re-verified for the jti to be able to be verified?

PR #421, opened 2021-09-24
Issue 420: CORE-ch 4.2 Data representation added. TRANSPORT-Read example with array.

UlfBj:


UlfBj:
https://raw.githack.com/UlfBj/automotive/gh-pages/spec/VISSv2_Core.html
https://raw.githack.com/UlfBj/automotive/gh-pages/spec/VISSv2_Transport.html

PR #427, opened 2021-10-20
Issue 369. MQTT added to the TRANSPORT specification.

UlfBj:


UlfBj:
Browser compatible version of the TRANSPORT document
https://rawcdn.githack.com/UlfBj/automotive/5128aeca1313bd5264968455faa6ca2e179fcb21/spec/VISSv2_Transport.html

spec/VISSv2_Transport.html
```diff
@@ -1241,6 +1280,67 @@ <h2>Unsubscribe</h2>
...
+
+        <section id="term-def">
+          <h2>Application Level Protocol</h2>
+          <p>
+            For MQTT to support the complete VISSv2 interface, as decribed in the "Interface" chapter of the CORE specification, 
+            an application level protocol that runs on top of MQTT is added. 
+            It is described in the following, please also see the sequence diagram below. 
+            To emulate the client-server pattern that is described in the CORE specification, the vehicle server, via its vehicle client, 
+            issues a subscribe request to the broker on a topic named VINXXX/Vehicle, where VINXXX is a unique identity of the vehicle, 
+            a "pseudo-VIN" generated by calculating the SHA256 hash of the vehicle VIN concatenated with the string "W3C-VISSv2",
```
erikbosch: Why do we need this pseudo-VIN handling? A regular VIN is 17 characters so why not use VIN directly?

If we are to keep the PseudoVIN handling then I think an example would be good, showing a real VIN and how the corresponding topic looks like

spec/VISSv2_Transport.html
```diff
@@ -1241,6 +1280,67 @@ <h2>Unsubscribe</h2>
...
+            It is described in the following, please also see the sequence diagram below. 
+            To emulate the client-server pattern that is described in the CORE specification, the vehicle server, via its vehicle client, 
+            issues a subscribe request to the broker on a topic named VINXXX/Vehicle, where VINXXX is a unique identity of the vehicle, 
+            a "pseudo-VIN" generated by calculating the SHA256 hash of the vehicle VIN concatenated with the string "W3C-VISSv2", 
+            then taking the most significant 32 bytes of its hex representation.<br>
+            The client on the "cloud side" of the broker is expected to have access to this pseudo-VIN. 
+            How it obtains it is out of scope for this specification. 
+            When the client wants to issue a request to the vehicle server it first generates a unique topic name, 
+            which it subscribes to at the broker. It then generates a JSON formatted payload with the general structure<br>
+            {"topic":"aUniqueTopic", "request":"aVissv2Request"}<br>
```
erikbosch: The picture use the term "ARequestTopic" and "VISSv2 request". I suggest that we align the names

spec/VISSv2_Transport.html
```diff
@@ -1241,6 +1280,67 @@ <h2>Unsubscribe</h2>
...
+            "request" key in the message to the vehicle server. When the vehicle client receives the response to this request, 
+            it publishes it to the broker associated with the topic name that was the string value of the "topic" key name in the 
+            message it previously received from the broker.<br>
+            The broker will then forward this message to the cloud side client that earlier subscribed to this topic name, 
+            which concludes the clent-server based request-response as described in the CORE specification.<br>
+            In the case of subscription requests the vehicle client needs to save the subscriptionId found in the subscribe response, 
+            together with the topic name associated to the subscribe request. When the vehicle server later issues notifications, 
+            the vehicle client must parse the subscriptionId from it, and retrieve the topic name associated to it. 
+            The client shall delete the saved topic name and subscriptionId when it receives an unsubscribe request in a message from the broker.<br>
```
erikbosch: As we have two clients I suggest to use the sentence with "The vehicle client ..."

spec/VISSv2_Transport.html
```diff
@@ -1241,6 +1280,67 @@ <h2>Unsubscribe</h2>
...
+            In following requests from the cloud side client, the client may reuse the unique topic name from the previous request-response cycle, 
+            or it may generate a new unique topic name. If a new topic name is generated, and unsubscribe should be issued on the old topic name.
```
erikbosch: Having difficulties understanding this sentence:

 "If a new topic name is generated, and unsubscribe should be issued on the old topic name."

spec/VISSv2_Transport.html
```diff
@@ -1241,6 +1280,67 @@ <h2>Unsubscribe</h2>
...
+            The broker will then forward this message to the cloud side client that earlier subscribed to this topic name, 
+            which concludes the clent-server based request-response as described in the CORE specification.<br>
+            In the case of subscription requests the vehicle client needs to save the subscriptionId found in the subscribe response, 
+            together with the topic name associated to the subscribe request. When the vehicle server later issues notifications, 
+            the vehicle client must parse the subscriptionId from it, and retrieve the topic name associated to it. 
+            The client shall delete the saved topic name and subscriptionId when it receives an unsubscribe request in a message from the broker.<br>
+            In following requests from the cloud side client, the client may reuse the unique topic name from the previous request-response cycle, 
+            or it may generate a new unique topic name. If a new topic name is generated, and unsubscribe should be issued on the old topic name. 
+            The vehicle client does not need to reissue its topic name.<br>
```
erikbosch: Do not really understand this sentence either

spec/VISSv2_Transport.html
```diff
@@ -1241,6 +1280,67 @@ <h2>Unsubscribe</h2>
...
+            {"topic":"aUniqueTopic", "request":"aVissv2Request"}<br>
+            where "aUniqueTopic" is the uniques topic name it just subscribed to, and "aVissv2Request" is the request for the vehicle server. 
+            This request MUST follow the payload format that is specified in the Websocket chapter of this specification.
+            This JSON message is then issued to the broker, associated to the topic VINXXX/Vehicle. 
+            This message will then be forwarded by the broker to the vehicle client, which forwards the string being the value of the 
+            "request" key in the message to the vehicle server. When the vehicle client receives the response to this request, 
+            it publishes it to the broker associated with the topic name that was the string value of the "topic" key name in the 
+            message it previously received from the broker.<br>
+            The broker will then forward this message to the cloud side client that earlier subscribed to this topic name, 
+            which concludes the clent-server based request-response as described in the CORE specification.<br>
```
erikbosch: Change "clent" to "client"

spec/VISSv2_Transport.html
```diff
@@ -1241,6 +1280,67 @@ <h2>Unsubscribe</h2>
...
+            In the case of subscription requests the vehicle client needs to save the subscriptionId found in the subscribe response, 
+            together with the topic name associated to the subscribe request. When the vehicle server later issues notifications, 
+            the vehicle client must parse the subscriptionId from it, and retrieve the topic name associated to it. 
+            The client shall delete the saved topic name and subscriptionId when it receives an unsubscribe request in a message from the broker.<br>
+            In following requests from the cloud side client, the client may reuse the unique topic name from the previous request-response cycle, 
+            or it may generate a new unique topic name. If a new topic name is generated, and unsubscribe should be issued on the old topic name. 
+            The vehicle client does not need to reissue its topic name.<br>
+            The payload format of the responses/notifications that the client receives follow the payload format that is specified 
+            in the Websocket chapter of this specification.
+          <img src="images/mqtt_vissv2_protocol.jpg" alt="VISSv2 over MQTT.">
```
erikbosch, line 1324: What is the W3C approach to images? Do we want the source to be present in the repo as well, so that anyone can update it if needed?

Another comment on the image: Maybe we shall say a few words on the communication between "Vehicle Client" and "VISSv2 Server". Is this implementation dependent or do we expect it to be using e.g. Websocket or https.

spec/VISSv2_Transport.html
```diff
@@ -1241,6 +1280,67 @@ <h2>Unsubscribe</h2>
...
+          </p>
+        </section>
+
+        <section id="term-def">
+          <h2>Security Aspects</h2>
```
erikbosch, line 1329: Do we need to say something about authentication/authorization. Assume e.g. that there are multiple cloud clients that want to subscribe to information on "Door/IsOpen", but only one of them are allowed to do it. Will the "Vehicle Client" then subscribe to the "VISSv2 server" multiple times, one time for each cloud client using the token of that particular "cloud client"?

spec/VISSv2_Transport.html
UlfBj: Privacy reason was the idea behind it.

spec/VISSv2_Transport.html
UlfBj: "aUniqueTopic" represents the topic name, while "aVissv2Request" represents the actual VISSv2 request. 
So they cannot be the same.

spec/VISSv2_Transport.html
UlfBj: Agree.

spec/VISSv2_Transport.html
UlfBj: Typo. and -> an

spec/VISSv2_Transport.html
UlfBj: Please specify what is unclear.

spec/VISSv2_Transport.html
UlfBj: Done

spec/VISSv2_Transport.html
UlfBj: I think it is an implementation issue. The requirement on it is that it must be able to transport the JSON message payload.

spec/VISSv2_Transport.html
erikbosch: I think you missunderstood me.

I mean, on the line above we use the term "aUniqueTopic", in the picture we use the term "aRequestTopic". Isn't "aUniqueTopic" and "aRequestTopic" the same thing? I.e. if we use "aUniqueTopic" in text thenj we shall also use it in the picture instead of "aRequestTopic"?

Similarly, on the line above we say "aVissv2Request" but in the picture "VISSv2 request".

spec/VISSv2_Transport.html
erikbosch: I think the privacy topic was briefly discussed in a previous meeting. But at least in Sweden VIN does not seem to be that sensitive, if you know the license plate number of a car you can easily get the VIN.

spec/VISSv2_Transport.html
erikbosch: It is "(re)issue a topic name" that puzzles me. What is it more exactly that the client does not need to do again?

spec/VISSv2_Transport.html
UlfBj: The VISSv2 access control can be used also over the MQTT transport. I think it should be mentioned in this context. 
Subscriptions are processed per client.

UlfBj:
Updated readable version per comments above.
https://rawcdn.githack.com/UlfBj/automotive/77032393cf469d35e1554035d30827b1a640a914/spec/VISSv2_Transport.html

SebastianSchildt:
I feel there are two options



1. 
 Use the VIN directly, not "obfuscation", hashing. Argument: Access control you need on your broker anyway, and VIN is not "privacy" critical. You would now how to access a specific vehicle
 
2.
Just define an "identifier" (string? uuid?), and leave it up to the application/deployment:  Argument: The person deploying VISS would now "what is what" (OEM might connect to VINs, fleet operator might have internal car numbers...). Raw VIN might still be an option. You loose "discoverability" based on VIN


Both work, I feel 2 is better. 


"Mangling"/"Obfuuscating" the VIN in a standardised  manner without random element so that 1:1 lookup table can be created just is effort in servers and clients, and provides not advantage over 1.

spec/VISSv2_Transport.html
peterMelco: VIN is absolutely sensitive in terms of privacy related to data collection.

erikbosch:
Concerning the VIN/Hash discussion - in the "core" document for tokens we say

`VIN: The vehicle identification number. This may be a pseudo-VIN, or any other identity that uniquely links to the vehicle in the access control ecosystem.`

Would the same (or similar) definition work here as well, instead of stating an explicit method (SHA256 Hash of VIN)? I assume that the "Vehicle Client" and/or "VISSv2 server" anyway need to be able to do a reverse lookup so it can update the right vehicle, and/or check that the used token gives rights to access data on this particular vehicle.

UlfBj:
PR updated, see commit e159178 above.
Link below shows latest version.
https://rawcdn.githack.com/UlfBj/automotive/e1591783f88f3f4e0c17d252f9341c991aef6f2f/spec/VISSv2_Transport.html

tguild:
Since VIN is Vehicle IDentification Number, we thought on call it might be best to clarify there is potentially no relation between VID and VIN.

That could be done by following up on this statement:

where VID is an identity that uniquely links to the vehicle in the access control ecosystem.

with:

This vehicle identity is not necessarily the manufacturer's Vehicle Identification Number (VIN).

PR #431, opened 2021-11-11
Vehicle identity claim optional in AGT and AT.

UlfBj:


UlfBj:
https://rawcdn.githack.com/UlfBj/automotive/cb6fda7c3942448ff05a2705b66c0486017cfa2a/spec/VISSv2_Core.html

spec/VISSv2_Core.html
```diff
@@ -1222,16 +1308,19 @@ <h2>Access Token</h2>
...
           }
 	  </pre>
-	  Algorithm (alg) shall be set to any valid algorithms according to [[RFC7518]].<br>
-	  Type (typ) shall be set to "JWT".<br>
-          Issued at (iat) shall be set to the time of token issuance, in Unix time.<br>
-          Expiry (exp) shall be set to the time when the token expires, in Unix time.<br>
-          The scope (scp) shall be set to the <a href="#purpose-list-def">purpose list</a> short name, 
+	  The algorithm (alg) claim shall be set to any valid algorithms according to [[RFC7518]].<br>
+	  The type (typ) claim shall be set to "JWT".<br>
+	  The vehicle identity claim (vin) may be present or not, depending on its presence in the
```
erikbosch: Must it always be so that if VIN is present in the AGT, then it must also be present in the AT? Or is that independent?

I assume that ATS to be able to issue a vehicle-specific AT must have a vin in the AGT, that all AGTs issued by AGT are signed with the same key. But we have two mechanisms to make ATs vehicle-specific:

- Include an explicit vin claim in the AT
- Sign with a vehicle specific key, then signature check will fail if trying to use the token toward the "wrong" vehicle.

If so, having vin in AGT does not necessarily imply that there will be a vin in AT, even if could be a good practice, if nothing else so for verifying that you use the right token. Shoul we possibly rephrase to something like below?
 
`The vehicle identity claim (vin) may be present or not, depending on its presence in the access grant token and whether vehicle specific signatures are used or not.`

UlfBj:
>> Must it always be so that if VIN is present in the AGT, then it must also be present in the AT? 

I think it is a reasonable requirement. It may lead to a scenario where the signature validation is an implicit validation of that the token is used at the correct vehicle, and thus the VIN validation is redundant. But it also keeps the door closed for the case that no such validation is done. A simplified logic with the potential cost of a redundant validation.

spec/VISSv2_Core.html
```diff
@@ -1020,37 +1068,50 @@ <h2>Access Token Server</h2>
...
+        <a href="#proof-of-possession-def">proof of possession</a> using the public key contained in the token.<br>
+        The <a href="#access-grant-token-def">access grant token</a> signature validation is done with the 
+        <a href="#access-grant-server-def">access grant server</a> public key. 
         <br>
-        To allow for some time synchronization inaccuracy and minor network latency, expiry times should be set to include potential modest margin of error, possibly as long as tens of seconds.<br>
-        If the <a href="#client-context-def">client context</a> as declared in the <a href="#access-grant-token-def">access grant token</a> does not match in the set of allowed roles for the requested <a>purpose</a> in the <a href="#purpose-list-def">purpose list</a>, 
+        To allow for some time synchronization inaccuracy and minor network latency, 
+        expiry times should be set to include potential modest margin of error, possibly as long as tens of seconds.<br>
+	If a vehicle identity claim (vin) is present in the <a href="#access-grant-token-def">access grant token</a>,
```
erikbosch: Tabs instead of whitespace on line 1078 and 1079?

spec/VISSv2_Core.html
```diff
@@ -1066,13 +1127,22 @@ <h2>Access Control Server</h2>
...
+          <a href="#access-grant-token-def">access grant token</a> first, 
+          before revisiting the <a href="#access-token-server-def">access token server</a>.<br>
+          The server SHOULD support caching of a limited number of <a href="#access-token-def">access tokens</a>. 
+          The <a href="#access-token-def">access token</a> MUST be included in the cache after a first successful request and 
+          MUST be removed once they expire. 
 	  The client can decide to include only the jti claim in following requests in order to save bandwith. 
-	  At any point the server might decide to remove any token from the cache. In this case the client will get a "401, missing_token" error and will be forced to send the whole <a href="#access-token-def">access token</a> again.
-          For <a href="#client-def">client</a> requests that are not granted due to access control, the VISSv2 server MUST return one of the error codes shown in the table below.<br>
+	  At any point the server might decide to remove any token from the cache. In this case the client will get a "401,
```
erikbosch: Also tabs here (line 1142+1143)

erikbosch (approved):
Some minor editorial things like there seems to be tabs instead of whitespace in a few places, but looks in general ok for me

PR #435, opened 2021-11-23
open flow alternative

isaacagudo:
I tried to minimize the changes required to include explicit signals in the access token. Only relevant sections are viss server and access token.

### DIFF
--- a/spec/VISSv2_Core.html
+++ b/spec/VISSv2_Core.html
@@ -1118,8 +1118,6 @@
           <ul>
             <li>Token signature.</li>
             <li>Token expiry time.</li>
-            <li>Token signal scope.</li>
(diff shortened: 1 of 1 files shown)


UlfBj:
https://rawcdn.githack.com/w3c/automotive/25dbc376ee83527b43b8efce63eaadefa752e3a0/spec/VISSv2_Core.html#access-token

UlfBj:
Change
"Except for the vehicle identity (vin) claim that is optional, the Access token SHALL have the following claims in header and payload."
to
"Except for the vehicle identity (vin), and client context (clx) claims that are optional, the Access token SHALL have the following claims in header and payload."

Add to clx description: If the scope claim is set to a purpose, the client context claim MUST be present in the token.

PR #437, opened 2021-12-03
Issue381/385: Optional server capabilities.

UlfBj:


UlfBj:
https://rawcdn.githack.com/UlfBj/automotive/fcd480bd98838e7c09a16008e6e64cb629904f97/spec/VISSv2_Core.html
https://rawcdn.githack.com/UlfBj/automotive/009d0b309aa415953e9e052e0be23a958ac6ca41/spec/VISSv2_Transport.html

spec/VISSv2_Core.html
```diff
@@ -584,11 +584,11 @@ <h2>Filter Request</h2>
           The value data varies depending on the type, 
           is enclosed in one object, or optionally an array of two objects, and is described in the following chapters.</li>
           </ul>
+          The server MUST support the timebased and change types, the other types are optional.<br>
```
erikbosch, line 592: When we say "must support timebased", is it clear what that means? In the example in https://rawcdn.githack.com/UlfBj/automotive/009d0b309aa415953e9e052e0be23a958ac6ca41/spec/VISSv2_Transport.html#subscribe an interval of 3 milliseconds are requested. I assume most VISS_implementations will not be able to provide data every 3 milliseconds. Have we said what shall be returned if the server cannot support 3 ms as interval but could support 100 ms. Or if it cannot support 207 ms but could support 200 ms?

spec/VISSv2_Core.html
```diff
@@ -1650,5 +1657,93 @@ <h2>Access Control Selection</h2>
...
     <section id="tof" class="appendix"></section>
+
+    <section id="optional-server-capabilities" class="appendix">
+        <h2>Optional Server Capabilities</h2>
+        <p>
+        On receiving a dynamic-metadata request on optional capabilities the server SHALL respond with a JSON object that for each of the keys
+        in the list below contains the associated values in the list that the server supports. 
+        Both the keys and the values are represented as strings, if a key has more than one value, these are represented as an array, 
+        see example below.<br>
+        {"flt": ["paths", "clog"], "acc": ["sterm", "lterm", "open"], "trp": "wss"}
```
erikbosch: "open" not defined below. Should be "sset"?

spec/VISSv2_Transport.html
```diff
@@ -299,8 +299,10 @@ <h2>Authorization</h2>
 
     <section id="transport-protocols">
       <h2>Transport Protocols</h2>
-      <p>The transport protocols supported are the secure versions of the HTTP and WebSockets protocols. 
-      Further transport protocols may be supported in future versions of this specification.
+      <p>The transport protocols supported are the secure versions of the HTTP and WebSockets protocols, 
+      and the MQTT protocol, on which a thin application layer protocol is applied.<br>
+      The server MUST support the HTTP protocol, the other protocols are optional. <br>
```
erikbosch: previously we said "_The server MUST support the timebased and change types_"

but we also say:

_The types timebased, range, change, and curvelog are only applicable for subscription requests. Subscription requests are only supported by the websocket transport protocol._

Isn't that a bit contradicting. We say that Timebased/Change is mandatory, but they are require websocket which is optional.

spec/VISSv2_Core.html
erikbosch: Timebased and Change only required if MQTT or Websocket is supported, right. No need (well not possible) to support them if only HTTP supported.

spec/VISSv2_Core.html
```diff
@@ -1650,5 +1657,93 @@ <h2>Access Control Selection</h2>
...
+            <td>range</td>
+            <td>Range filter</td>
+          </tr>
+          <tr>
+            <td>clog</td>
+            <td>Curve logging filter</td>
+          </tr>
+          <tr>
+            <td>hist</td>
+            <td>History filter</td>
```
erikbosch: Concerning filters - do we see a need to be able specify/communicate different capabilities for different protocols. Maybe we support Range for MQTT but not for HTTP, and History for HTTP but not for MQTT

spec/VISSv2_Transport.html
UlfBj: Good point. Needs to be discussed.

spec/VISSv2_Core.html
UlfBj: It seems odd to me that the filter functionality should have a dependence to the transport protocols?

spec/VISSv2_Core.html
UlfBj: Fixed.

spec/VISSv2_Core.html
UlfBj: The 3 comes from when the time unit was seconds. It will be changed to 500. 
Your other comments should be discussed in the WG.

spec/VISSv2_Core.html
erikbosch: There are at least an implicit dependency, as some of the filters (like range) only are applicable for subscriptions, and subscription is only supported by websocket protocol (at stated in core, I assume MQTT shall be mentioned there as well)

_The types timebased, range, change, and curvelog are only applicable for subscription requests. Subscription requests are only supported by the websocket transport protocol._

UlfBj: >> The types timebased, range, change, and curvelog are only applicable for subscription requests. Subscription requests are only supported by the websocket transport protocol.

The sentence should mention also MQTT. I think it sufficiently states the filter capabilities vs transport protocols.

PR #440, opened 2022-01-04
Document improvements

UlfBj:


spec/VISSv2_Core.html
```diff
@@ -429,11 +429,13 @@ <h2><dfn>Subscription</dfn></h2>
...
         <h2>Error Message</h2>
         <p>The server MUST inform a client about errors ocurring in interactions between the two, whether it is in a synchronous 
-        error response, or an asynchronous error notification as a result of a previous subscribe.
+        error response, or an asynchronous error notification as a result of a previous subscribe. The error message has three arguments, 
+        of which subscriptionId is optional.
         </p>
           <p>Arguments: 
             <ul>
               <li><a>error message</a> <a href="#error-information"></a></li>
+              <li><a>subscriptionId</a> The subscriptionId MUST be present in error notifications.</li>
```
erikbosch, line 438: On line 433 we say subscriptionId is optional. Here we say that it is mandatory in some cases. Confused me for a while.

Should we possibly consider rephrasing on line 432/433 to something like:

_The error message has three arguments, of which subscriptionId is mandatory only for error notifications._

Then we do not need to repeat that information here

UlfBj: Accepted. Comes in next PR queued up after this.

wonsuk73: Regarding a  description style of arguments it might be better if we can make a sync of them. For example. "error message" line have a link for 5.2.1 error message but doesn't have a short descriptioin. In case of "timestamp" and  "subscriptionId" only have a description but doesn't have a link.

spec/VISSv2_Transport.html
```diff
@@ -511,20 +511,17 @@ <h2>Signal Discovery Read</h2>
...
+          <section id="https-dynamic-metadata-read">
+            <h2>Dynamic Metadata Read</h2>
             <p>
               The dynamic metadata, i. e. any other metadata kept by the vehicle system, 
               is retrieved by the setting the "type" to  "dynamic-metadata".
-              If the value is set to an empty string, then all dynamic metadata is returned. 
-              If only specific metadata is desired, 
-              the "value" can be set to the domain names as specified in the CORE: Server Capabilities, 
-              e. g. "value":["filter", "access_ctrl"].<br>
+              The value MUST be set to one of the domain names as specified in the CORE: Dynamic Metadata Filter Operation chapter.
```
wonsuk73, line 519: It might be good if we add the link to "the CORE: Dynamic Metadata Filter Operation chapter" part.

spec/VISSv2_Transport.html
```diff
@@ -1241,13 +1238,51 @@ <h2>Curve Logging Subscribe</h2>
...
                 "ts": "2020-04-15T13:37:00Z"
               }
                </code></pre>
             </p>
           </section>
+
+          <section id="range-subscribe">
+            <h2>Range Subscribe</h2>
+            <p>
+            Subscription to a range of values, that can be either unbounded, or bounded as in the example below.
```
wonsuk73, line 1252: How about revise like this. "Subscription to a range of values, that can be either unbounded, or bounded as in the example below. For a more information how to use range of values, refer the Range Filter Operation chapter in the Core documentation."

spec/VISSv2_Transport.html
```diff
@@ -1241,13 +1238,51 @@ <h2>Curve Logging Subscribe</h2>
...
+              {
+                "action": "subscription",
+                "subscriptionId": "12345",
+                “data”:{“path”: ”Vehicle/Drivetrain/FuelSystem/Level”, 
+                        “dp”:{“value”: ”51”, “ts”: ”2020-04-15T14:00:00Z”},
+                "ts": "2020-04-15T14:00:00Z"
+              }
+               </code></pre>
+            </p>
+          </section>
```
wonsuk73, line 1285: looks good to me!

@UlfBj Thanks for the PR! :)

spec/VISSv2_Transport.html
UlfBj: Your comment raises the general question how to cross reference between the CORE and TRANSPORT documents. Currently it is done with pure text as the case you point to above. 
These references are in several places in respective document, and I assume the URL will change when/if it will become a standard,so it would be preferred to have the reference at one place, such as in the References chapter? But then it would only point to the document as a whole, not into specific chapters. 
I am not sure how to best solve it, I think it needs a WG discussion.

spec/VISSv2_Transport.html
UlfBj: Updated as proposed.

spec/VISSv2_Transport.html
wonsuk73: @UlfBj Okay! I am also going to look for the way for cross reference between the CORE and TRANSPORT :)

UlfBj: @wonsuk73 Good:). 
I just created an issue on it, where we can continue the discussion.

wonsuk73: @UlfBj That's great! 👍

spec/VISSv2_Transport.html
wonsuk73: @UlfBj Thanks! :)

PR #452, opened 2022-04-07
Issue 450. Inverse bounded range support.

UlfBj:


UlfBj:
https://rawcdn.githack.com/UlfBj/automotive/5ebff572b796aa561cdfc0c77ed4fa00cd9b1fee/spec/VISSv2_Core.html

spec/VISSv2_Core.html
```diff
@@ -662,11 +662,22 @@ <h2>Time Based Filter Operation</h2>
...
-                The value may be an array of two of these objects, 
-                which will then be combined through a logical AND to support expressions of a bounded range. 
-                A range event is triggered whenever a signal is updated, and fulfilling the range expression.<br><br>
+                The range filter operation supports three types of ranges:<br>
+                1. Unbounded range<br>
+                2. Bounded range<br>
+                3. Inverse of bounded range<br>
+                The range operation issues notifications to the client when the logical expression defined in the "value" key-value pair evaluation is true, and the signal is updated.<br>
+                For the unbounded range, the "value" consists of one JSON object that contains the range boundary, and the logical operator for the boundary, 
+                {"boundary-op":"X", "boundary": "Y"},
```
erikbosch: It shall be `logic-op` right? At least that is what we use in all examples in transport:

```
{
                "action": "subscribe",
                "path": "Vehicle/Drivetrain/FuelSystem/Level",
                "filter": "filter":{"type":"range","value":[{"logic-op":"gt","boundary":"50"},{"logic-op":"lt","boundary":"55"}]},
                "requestId": "6578"
              }
```

UlfBj: With this PR there will now be two different "logic-op", therefore I renamed them to "boundary-op" and "combination-op", respectively. If we go with that, the examples need to be updated. Thanks for catching that.

erikbosch: As long as we use use the same term in core and transport I am happy :-)

UlfBj:
https://rawcdn.githack.com/UlfBj/automotive/a782f88461b89be9b430678e2c8347e32534b89e/spec/VISSv2_Core.html
https://rawcdn.githack.com/UlfBj/automotive/a782f88461b89be9b430678e2c8347e32534b89e/spec/VISSv2_Transport.html

spec/VISSv2_Core.html
```diff
@@ -678,14 +701,14 @@ <h2>Range Filter Operation</h2>
...
                 For boolean values the following expressions shall be supported:<br>
                 "value":{"logic-op":"gt", "diff": "0"} This leads to a trigger event when the value goes false->true.<br>
                 "value":{"logic-op":"lt", "diff": "0"} This leads to a trigger event when the value goes true->false.<br>
                 "value":{"logic-op":"ne", "diff": "0"} This leads to a trigger event when the value goes true->false OR false->true.
                 <br><br>
-                (**)The supported logical operators are ["eq", "ne", "gt", "gte", "lt", "lte"], 
+                (**)The supported logic operands are ["eq", "ne", "gt", "gte", "lt", "lte"], 
                 where "eq" is "equal", "ne" is "not equal", "gt" is "greater than", "gte" is "greater than or equal", 
                 "lt" is "less than", "lte" is "less than or equal".<br>
                 Examples can be found in the [[viss2-transport]] specification.
```
erikbosch, line 714: Changes looks good, but when I reviewed them I noticed that we say that examples on `logic-op` (with diff) exists in Transport-section. But it doesn't, right`? Remove line or add example?

UlfBj: Good catch. Example added in the Transport doc.
https://rawcdn.githack.com/UlfBj/automotive/e68eb41ec4036d036f9172d73a15f4a74eef781d/spec/VISSv2_Transport.html

erikbosch (approved):

PR #470, opened 2023-01-02
Issue 460: value changed to parameter.

UlfBj:


tguild:
discussed on call today, agreement on terminology change

PR #471, opened 2023-01-02
Issue 462:subscriptionNotification table updated with ts entry.

UlfBj:


spec/VISSv2_Transport.html
```diff
@@ -1087,7 +1087,7 @@ <h2>Subscribe</h2>
 	  <tr><th rowspan="3"><dfn data-dfn-type="dfn" id="dfn-subscriptionnotification">subscriptionNotification</dfn></th><td><a href="#dfn-action">action</a></td><td><a href="#action-def">Action</a></td><td>Yes</td></tr>
 	  <tr><td><a href="#dfn-subscriptionid">subscriptionId</a></td><td>string</td><td>Yes</td></tr>
 	  <tr><td><a href="#dfn-value">data</a></td><td>object/array</td><td>Yes</td></tr>
-	</tbody>
+	  <tr><td><a href="#dfn-ts">ts</a></td><td>string</td><td>Yes</td></tr>	</tbody>
```
erikbosch: Possibly looks nicer with a line break before `</tbody>`

UlfBj: You are so diplomatic in your comments Erik :)
It will be fixed.

PR #472, opened 2023-01-02
Issue 469: notification changed to event.

UlfBj:


tguild:
discussed on call today, agreement on terminology change

PR #480, opened 2023-04-11
Issue461:Support for complex datatypes.

UlfBj:


spec/VISSv2_Core.html
```diff
@@ -302,10 +302,22 @@ <h2>Data representation</h2>
...
+          A struct with the following declaration<br>
+          ```
+          struct {
+              field1 <datatype>
+              field2 <datatype>
+          }
+          ```
+          is represented by the following JSON object<br>
+          {"field1":"X", "field2":"Y"}<br>
+          where X and Y are the actual values of respective datatype, here represented for simplicity as string values.
```
erikbosch, line 319: For "plain" signals we say above that the value always shall be string values, i.e. "123", "true", "ABC" and so on.  Shall it be string values here as well or should it be "native" JSON representation? I.e.:

`{"my_int":123, "my_bool": true, "my_string": "my_value", "my_int_array":  [10, 20, 30]}`

or rather:

`{"my_int":"123", "my_bool": "true", "my_string": "my_value", "my_int_array":  ["10", "20", "30"]}`

Maybe rephrase to make it more clear.

Would we benefit from having more examples?

PR #484, opened 2023-07-17
Issue 482:Token handle

UlfBj:


UlfBj:
A rendered version is found here:
https://rawcdn.githack.com/UlfBj/automotive/9086a4a152312a8f0ac8d0d764c5a5b84df7c26e/spec/VISSv2_Core.html

UlfBj:
Updated version:
https://rawcdn.githack.com/UlfBj/automotive/4dc39ce9e8a2e7c751b9fae43f19a08fde93c393/spec/VISSv2_Core.html

tguild:
@isaacagudo discussing this with Ulf, made sense to me and we want your thoughts.

isaacagudo:
I think it is ok, better not to go into specifics regarding how the handle is generated from the token. I would then propose to change this text
`If an access token is cached then the server shall copy the signature part of the JWT formatted access token and return that in the response to the client. The server may truncate the signature in order to reduce the size of the handle, but it must be at least 24 bytes long. ` 
keeping it as general as possible, e.g. 
`If an access token is cached then the server shall return a token handle of at least 24 bytes long. `

